PR #3, opened 2016-03-24
Make our VCL configuration more consistent

mattbostock:
Our timeouts when connecting to the origin are different in our `www`,
`assets` and `performanceplatform*` services.

Make them consistent, using these values:

    first_byte_timeout = 15s
    between_bytes_timeout = 10s

These seem more appropriate than the previous values of 120s, used for
for both timeouts, in the `www` service:

- 15 seconds is long enough to make a user wait (referring to the first
  byte timeout); in that case it's probably best to failover to the
  mirror

- Nginx, which we use at origin, already buffers proxied requests from
  upstreams, so it's unlikely that more than 10 seconds would pass
  between bytes for reasons other than networking:
  http://nginx.org/en/docs/http/ngx_http_proxy_module.html#proxy_request_buffering

Also set the probe interval to 10 seconds, which seems more sensible
than the previous interval of 60 seconds used by the `assets` service.
This should allow us to detect when origin is unhealthy more quickly.
We did the same for the `www` service in 574e48b.

* * *

Also, make the `host` and `share_key` parameters mandatory in the assets service.

### DIFF
--- a/vcl_templates/apt.vcl.erb
+++ b/vcl_templates/apt.vcl.erb
@@ -3,16 +3,16 @@ backend F_apt {
     .connect_timeout = 1s;
     .dynamic = true;
     .port = "<%= config.fetch('apt_port', 80) %>";
-    .host = "<%= config['apt_hostname'] %>";
+    .host = "<%= config.fetch('apt_hostname') %>";
     .first_byte_timeout = 15s;
     .max_connections = 200;
     .between_bytes_timeout = 10s;
-    .share_key = "<%= config['service_id'] %>";
+    .share_key = "<%= config.fetch('service_id') %>";
 
     .probe = {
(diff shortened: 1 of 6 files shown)


deanwilson:
Fetch and 'normal' hash element retrieval have different failure modes if the key is not found. This change will cause it to return a nil rather than throw a key exception. Is that intended?

mattbostock:
@deanwilson My bad, will amend.

mattbostock:
@deanwilson Updated via 5773115.

deanwilson:
Can you think of any situations where we'd currently set a value to empty in the config file? If not then merge away.

mattbostock:
@deanwilson I don't think we ever would intentionally in these cases.

PR #4, opened 2016-04-29
How the cdn configs can be deployed

rjw1:
Update README to give a hint as to how the configs can be deployed.


deanwilson:
Merge when ready.

PR #6, opened 2016-05-20
Pre-empt request changes in vcl_recv before purge

ajlanghorn:
The previous logic for accepting a purge only from a predefined whitelist of
IP addresses was potentially flawed in that vcl_recv might further modify an
acceptable FASTLYPURGE request before it is performed at the edge, yet the
current logic would return a lookup immediately after this logic had
applied. This meant that if the request was ever changed in vcl_recv, it
would be returned instantly rather than the remainder of logic in the VCL
being performed.

Fixes this in order to futureproof vcl_recv to some extent, as well as neaten up
the purge conditional in to fewer lines.

### DIFF
--- a/vcl_templates/assets.vcl.erb
+++ b/vcl_templates/assets.vcl.erb
@@ -141,11 +141,9 @@ acl purge_ip_whitelist {
 sub vcl_recv {
 
   # Allow FASTLYPURGE from IPs defined in the ACL only, else return a HTTP 403
-  if (req.request == "FASTLYPURGE") {
-    if (client.ip ~ purge_ip_whitelist) {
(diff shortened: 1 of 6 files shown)


alexmuller:
Does this work? There's a `lookup` action happening which has been removed in the diff.

ajlanghorn:
@alexmuller I removed it after reading https://community.fastly.com/t/set-up-a-purge-acl/54/2. It's removed because if you have any other actions occurring below the `lookup`, then they're ignored, because the `lookup` has already occurred and the response has been returned. I'm using this way of doing things in my own VCL, and it seems to work fine... but YMMV, I suppose.

PR #7, opened 2016-05-20
Maintain one list of Fastly cache node IPs

ajlanghorn:
Rather than maintain a list of Fastly cache node IP addresses in multiple
VCL files, maintain them in one and drag them in using Varnish includes.

This PR is a bit work-in-progress, because I've nowhere to test these out,
but they should work fine. Some testing your side would be very kind :)

### DIFF
--- a/vcl_templates/assets.vcl.erb
+++ b/vcl_templates/assets.vcl.erb
@@ -1,3 +1,5 @@
+include purge_ip_whitelist
+
 backend F_origin {
     .connect_timeout = 5s;
     .dynamic = true;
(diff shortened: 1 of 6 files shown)


ajlanghorn:
To do:
- [ ] Account for mirrors-only and stack-and-mirrors purge whitelist behaviour

alexmuller:
I'm not sure I understand how this approach could work. The configure script renders the ERB and then sends a POST request to the Fastly API. The POST request will contain the `include` line you've added but there's no code that will upload the contents of the whitelist file.

ajlanghorn:
@alexmuller Indeed; I didn't make the relevant change to fastly-configure, because I wasn't sure if you were running deployments using it (a la jenkins.sh), or still had a copy on github.gds that you were still using.

alexmuller:
But is there any change to fastly-configure that could make this work? The VCL that we upload to the provider will contain a reference to a file that doesn't exist on the provider's machines (and there's no way we can guarantee where it will be on their filesystem).

ajlanghorn:
I believe you can pass `include`s to the API; I'm sure I've spoken with Austin about it before, but it must have been in my old GDS inbox and not in my personal one. So, unless the API has deprecated this, it should still work. I can't quickly find any documentation, but the additional overhead wasn't too complicated to do it inside fastly-configure.

I'll ask Fastly, since I have to speak to them this morning anyway :)

ajlanghorn:
I spoke to Justin at Fastly. Soon, this idea of keeping an ACL outside of a configuration version will be a first-class citizen in the API, but right now the `include` method is documented at https://docs.fastly.com/guides/vcl/uploading-custom-vcl#things-to-know-when-uploading-multiple-files.

alexmuller:
Would it not make more sense to render this as ERB rather than relying on functionality that's probably specific to one provider?

ajlanghorn:
`include` isn't, as I understand it, specific to one provider -- it's a built-in inside Varnish -- hence why I liked this solution in the first place. Rendering the ERB is an alternative that I considered, but I don't see a big advantage either way for a greenfield project. That being said, you do already have logic in your ERB templates, so maybe keeping them all the same is a good way of going. Up to you penguins, really.

PR #11, opened 2017-01-04
Add configuration for service.gov.uk

afda16:
Add configuration template for service.gov.uk to replace the current AWS
implementation with Fastly:
- Redirect HTTP for service.gov.uk to HTTPS service.gov.uk
- Redirect HTTPS service.gov.uk to www.gov.uk
- Set HSTS header on HTTPS reponse


alexmuller:
I think this code looks great 👍 

One nitpicky naming thing: would you mind renaming the file either `servicegovuk` or `servicedomain` or something like that? `service` by itself is a bit generic.

afda16:
Sure, renamed

PR #12, opened 2017-01-10
Treat any non-200 from a mirror as a 503

SamLR:
In order to improve UX responses from the mirror should either return a 200 status code (found) or a 503 (service not available). This is because there are many URLs that the mirrors do not cache that are valid (for example, `/search`). In these cases whilst it is technically correct to return a 404 this is likely to confuse a user who would assume they have mis-typed the URL rather than, as is the case, hit the website during an outage. The user response we would like to encourage in this case is retrying in a little while once the site has returned.

To return 503s for non-200 responses we have changed the varnish configuration for fastly so that any response from a mirror which isn't a 200 has its status code set to 503.

### DIFF
--- a/vcl_templates/www.vcl.erb
+++ b/vcl_templates/www.vcl.erb
@@ -233,6 +233,14 @@ sub vcl_fetch {
     return (pass);
   }
 
+  # The only valid status from our mirrors is a 200. They cannot return e.g.
+  # a 301 status code. All errors from the mirrors are set to 503 as they
+  # cannot know whether or not a page actually exists (e.g. /search is a valid
+  # URL but the mirror cannot return it).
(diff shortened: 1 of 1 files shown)


rjw1:
looks fine to me i do wonder which 503 it will serve though.

alexmuller:
Yeah, that's a good question. My assumption is that it will take the same backend response text (which in this case is the mirror error page) and only modify the status code.

rjw1:
@alexmuller that would be the case currently but what about when we are using s3?

alexmuller:
I think in that case S3 has the error page set which we could specify to be our standard error and that would end up propagated all the way through.

surminus:
Also think it looks fine and having the name of the backend has served to be extra useful!

PR #13, opened 2017-01-10
Configure A/B testing cookies and headers

suzannehamilton:
Add Varnish config specifically for the 'frontend' app, and configure it to set cookies and headers for an A/B test.

The A/B testing approach follows https://www.fastly.com/blog/ab-testing-edge, and uses a mixture of cookies and custom headers:

- Users are placed into A/B buckets in the Varnish config, rather than in the app. This means that a user can visit the site for the first time and be assigned an A/B testing cookie without their request necessarily hitting the app server, because the response comes from the cache and the cookie was added by Varnish
- The cookie is transformed into a custom header in Varnish, and we vary by that header. This means that the CDN and the govuk Varnish just cache two versions of the page - one for each A/B test variant. This avoids varying by cookie, which would cause Varnish to cache a version for each user.

I've split this into two commits to separate the boilerplate from the A/B config changes. The boilerplate is quite extensive, unfortunately - I copied it from the [Fastly docs](https://docs.fastly.com/guides/vcl/mixing-and-matching-fastly-vcl-with-custom-vcl), and it seems consistent with some of the other apps' custom Varnish config.

@carvil, @tijmenb and @alexmuller - could you take a look to see whether you agree with the approach and the implementation?

https://trello.com/c/QyDbSiR9/289-allow-a-b-testing-cookies-to-be-set

### DIFF
--- a/vcl_templates/www.vcl.erb
+++ b/vcl_templates/www.vcl.erb
@@ -198,6 +198,17 @@ sub vcl_recv {
     return(pass);
   }
 
+  if (!req.http.Cookie:ABTest-Example) {
+    if (randombool(5,10)) {
+       set req.http.X-ABTest-Example = "A";
+    } else {
+       set req.http.X-ABTest-Example = "B";
+    }
+  } else {
+    # Set the value of the header to whatever decision was previously made
+    set req.http.X-ABTest-Example = req.http.Cookie:ABTest-Example;
+  }
+
   return(lookup);
 }
 
@@ -262,6 +273,13 @@ sub vcl_miss {
 }
 
 sub vcl_deliver {
+  # Set the A/B cookie
+  if (req.http.Cookie !~ "ABTest-Example") {
+    # Set a fairly short cookie expiry because this is just an A/B test demo.
+    # We should choose a longer expiry for a real A/B test.
+    add resp.http.Set-Cookie = "ABTest-Example=" req.http.X-ABTest-Example "; expires=" now + 1d;
+  }
+
 #FASTLY deliver
 }
 


nickcolley:
Does this mean the first request will always be identical, and only repeat visits can be tested?

suzannehamilton:
Not quite sure what you mean, sorry. (Probably because I've been staring at Varnish config files for a while.) Are you asking whether the first request by a user always returns the same A/B variant because they don't yet have a cookie set?

vcl_templates/frontend.vcl.erb
```diff
@@ -0,0 +1,96 @@
...
+}
+
+sub vcl_miss {
+#FASTLY miss
+  return(fetch);
+}
+
+sub vcl_deliver {
+  # Set the A/B cookie
+  if (req.http.Cookie !~ "ABTest-Example") {
```
suzannehamilton: I've just noticed that this won't currently work because the govuk Varnish will strip out the original cookie from the request.

I've got a PR open for puppet (alphagov/govuk-puppet#5331) which will allow these cookies through, but I'd hoped that we could avoid making that change if we make the changes in _this_ PR.

An alternative would be to add an extra custom header to the request in `vcl_recv` which marks the request as new if it doesn't contain the A/B cookie. Then at this point in `vcl_deliver`, we only add the `Set-Cookie` header if the request was new. It's a bit non-standard, but the changes would be isolated to this file (the rest of the stack wouldn't care about the extra header), and I think it would be easier to maintain than the Varnish cookie filtering config. Thoughts?

carvil: Sounds sensible to me. I would also rather only have changes in one place to make it simpler. Since we only need to set the cookie once, an extra header would work.

nickcolley:
@suzannehamilton Yeah, I'm just wondering if there's a way we can consider initial visits. Potentially by serving a variant by default then setting a cookie? Might be mistaken.

tijmenb:
@nickcolley this solution does work like you would expect. The first time someone visits a page, we assign a random A/B bucket, set the cookie and show A or B. The next time we read the cookie and show A or B.

vcl_templates/frontend.vcl.erb
suzannehamilton: OK, I've updated `vcl_recv` and `vcl_deliver` with the new header.

suzannehamilton: I've changed my mind about this - I think the original code was fine. The govuk Varnish modifies the request that it passes on to the app server, but it doesn't affect the _original_ version of the request that the CDN inspects when processing the response.

suzannehamilton:
@tijmenb: As discussed, I've moved the A/B cookie config into `www.vcl.erb` and deleted `frontend.vcl.erb`.

suzannehamilton:
We've tested this in a temporary deployment to staging, and it all worked. 💃 🎉 🎆 

I'll merge this into master now.

alexmuller:
This looks really cool 👍 

Minor nitpick: [RFC 6648](https://tools.ietf.org/html/rfc6648) has deprecated the use of an `X-` prefix for HTTP headers so I don't think we need that. We should probably prefix with `GOVUK-` or similar instead.

suzannehamilton:
Thanks, that's a good point! We'll make that change tomorrow.

vcl_templates/www.vcl.erb
```diff
@@ -198,6 +198,17 @@ sub vcl_recv {
     return(pass);
   }
 
+  if (!req.http.Cookie:ABTest-Example) {
```
tijmenb: As discussed with @alexmuller, change this into: `if (req.http.Cookie !~ "ABTest-Example") {`

PR #16, opened 2017-01-26
Add gov.uk HSTS redirect

surminus:
Similar to service.gov.uk but without the inclusion of subdomains.

Named "hsts-govuk" as I thought just calling it was "govuk" might be a little confusing.

### DIFF
--- a/vcl_templates/tldredirect.vcl.erb
+++ b/vcl_templates/tldredirect.vcl.erb
@@ -9,6 +9,13 @@ backend dummy {
 }
 
 sub vcl_recv {
+  # Force SSL.
+  if (!req.http.Fastly-SSL) {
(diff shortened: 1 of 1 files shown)


surminus:
@alexmuller I spoke with @afda16 who suggested the redirect may be doing the wrong thing, can I confirm the expected behaviour with you?

alexmuller:
I think what you've got here is correct, but we've already got a service for this in Fastly - it's the `tldredirect.vcl.erb` file in this repo.

surminus:
Ooh, I see what you mean.

surminus:
This is repeated in https://github.com/alphagov/govuk-cdn-config/pull/20 so I'll close my PR.

PR #17, opened 2017-01-26
Configure an A/B test for the new education navigation

suzannehamilton:
- Invert the A/B split logic to improve readability. This change flips the A/B fraction calculation so it calculates the fraction of users in the B group, to be consistent with how we talk about A/B tests.
- Split users into two groups and assign them a cookie to keep them in that group. This determines whether a user sees the existing navigation for education or the new navigation.
  - For now, set the 'B' fraction to 0. This lets developers and remote
    testers access the 'B' variant without releasing it to the general
    public yet.
  - For now, configure a short cookie expiry time (1 day) so that we don't end up with users stuck seeing the 'A' version when the A/B test starts.

Trello: https://trello.com/c/j5X1BZJU/339-configure-cdn-for-navigation-a-b-test

I'll add a reminder to Trello so that we remember to increase the expiry time when we roll out the A/B test to the general public.

### DIFF
--- a/vcl_templates/www.vcl.erb
+++ b/vcl_templates/www.vcl.erb
@@ -212,6 +212,23 @@ sub vcl_recv {
     }
   }
 
+  if (req.http.User-Agent ~ "^GOV\.UK Crawler Worker" {
+    set req.http.GOVUK-ABTest-EducationNavigation = "A";
+  } else if (req.http.Cookie ~ "ABTest-EducationNavigation") {
+    # Set the value of the header to whatever decision was previously made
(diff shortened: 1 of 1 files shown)


suzannehamilton:
Marking as DO NOT MERGE because we realise we don't actually need to roll this out yet - we're adding a feature flag to hide the new navigation in production, and the remote test of the new nav isn't until the end of March.

tijmenb:
Can you split out the inversion & comment updates into a new PR so that we can merge that? That way this PR can apply cleanly and we have something to point to to show people how to run a A/B test.

suzannehamilton:
@tijmenb Done: #19 

When that's merged, I'll rebase this branch.

suzannehamilton:
Replaced by #29.

PR #21, opened 2017-01-31
Ensure the GOV.UK mirror stores the original version of the A/B test

suzannehamilton:
Identify requests from the mirror's crawler worker, and set the request header such that it only ever sees the original 'A' version of the example A/B test.

This is in preparation for the real A/B test of education navigation. It will affect many pages across the site, so it's important to make sure that the mirror contains a self-consistent version of all the affected pages.

Trello: https://trello.com/c/0v9u9TOs/353-ensure-that-the-gov-uk-mirror-only-mirrors-the-a-version-of-all-tests

The user-agent string comes from govuk_crawler_worker: https://github.com/alphagov/govuk_crawler_worker/blob/master/http_crawler/crawler.go#L71

### DIFF
--- a/vcl_templates/www.vcl.erb
+++ b/vcl_templates/www.vcl.erb
@@ -198,15 +198,18 @@ sub vcl_recv {
     return(pass);
   }
 
-  if (req.http.Cookie !~ "ABTest-Example") {
+  # Make sure the original 'A' version is stored in the GOV.UK Mirror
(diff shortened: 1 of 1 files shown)


suzannehamilton:
@alexmuller - any thoughts on this from an infrastructure perspective?

alexmuller:
This is significantly more logic and more change than we've ever had with the CDN config before, which makes me a little uncomfortable. But I understand why we're doing this and I think it's ok 👍

PR #22, opened 2017-02-02
[DO NOT MERGE] Set a unique user ID as part of a Javascript detection experiment

mobaig:
We're re-running the experiment detailed in
https://gds.blog.gov.uk/2013/10/21/how-many-people-are-missing-out-on-javascript-enhancement/.

One of the changes we're making this time is to identify users rather
than requests. This commit sets the cookie value if it's absent and
writes it to a request header to preserve the value across requests
(after it's been stripped by Varnish, specifically).

[DO NOT MERGE] - we need to update the cookie help page before we start setting this.

### DIFF
--- a/vcl_templates/www.vcl.erb
+++ b/vcl_templates/www.vcl.erb
@@ -306,7 +306,7 @@ sub vcl_deliver {
 
   # Set the Javascript detection cookie
   if (req.http.User-Agent !~ "^GOV\.UK Crawler Worker" && req.http.Cookie !~ "JS-Detection") {
-    add resp.http.Set-Cookie = "JS-Detection=" req.http.GOVUK-JS-Detection "; expires=" now + 5w;
+    add resp.http.Set-Cookie = "JS-Detection=" req.http.GOVUK-JS-Detection "; expires=" now + 5w "; path=/";
   }
 
 #FASTLY deliver
--- a/vcl_templates/www.vcl.erb
+++ b/vcl_templates/www.vcl.erb
@@ -215,8 +215,8 @@ sub vcl_recv {
   if (req.http.Cookie ~ "JS-Detection") {
     set req.http.GOVUK-JS-Detection = req.http.Cookie:JS-Detection;
   } else {
-  # If the identifier cookie isn't present, set the header to a unique value),
-  # dervived as per the suggestion in https://community.fastly.com/t/unique-request-identifier/468/2
+  # If the identifier cookie isn't present, set the header to a unique value,
+  # derived as per the suggestion in https://community.fastly.com/t/unique-request-identifier/468/2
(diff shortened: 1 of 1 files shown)


mobaig:
@suzannehamilton Would appreciate a review when you've got some time, have basically just ripped off the stuff you wrote for A/B testing!

cc: @tijmenb @carolinegreen -

I'm calling the cookie `JS-Detection` and have set an expiry of 5 weeks. Picked that latter number based on us wanting to run this experiment for about a month. Feedback welcome.

vcl_templates/www.vcl.erb
```diff
@@ -296,6 +304,11 @@ sub vcl_deliver {
     add resp.http.Set-Cookie = "ABTest-Example=" req.http.GOVUK-ABTest-Example "; expires=" now + 1d;
   }
 
+  # Set the Javascript detection cookie
+  if (req.http.User-Agent !~ "^GOV\.UK Crawler Worker" && req.http.Cookie !~ "JS-Detection") {
+    add resp.http.Set-Cookie = "JS-Detection=" req.http.JS-Detection "; expires=" now + 5w;
```
suzannehamilton: You'll probably need to add `"; path=/"` to the end of the cookie. Otherwise the browser defaults to the path of the individual page, and you'll end up with users with multiple IDs for different parts of the site.

We don't add this to the `ABTest-Example` cookie because it's only used on one page, but I've added it to `ABTest-EducationNavigation` in alphagov/govuk-cdn-config#17.

vcl_templates/www.vcl.erb
```diff
@@ -212,6 +212,14 @@ sub vcl_recv {
     }
   }
 
+  if (req.http.Cookie ~ "JS-Detection") {
+    set req.http.GOVUK-JS-Detection = req.http.Cookie:JS-Detection;
+  } else {
+  # If the identifier cookie isn't present, set the header to a unique value),
+  # dervived as per the suggestion in https://community.fastly.com/t/unique-request-identifier/468/2
```
suzannehamilton: Typo: dervived

suzannehamilton:
The config looks like it will do what you need it to. 👍 Just one suggestion on the cookie path.

CC @alexmuller, since he's been reviewing my changes to the config.

vcl_templates/www.vcl.erb
mobaig: 🤦‍♂️ 

fixed!

alexmuller:
Looks like there are references to cookies named both `GOVUK-JS-Detection` and `JS-Detection` here.

mobaig:
@alexmuller confusingly, it's actually one of the header names I got wrong. Just pushed a correction.

So what we've got is a header `GOVUK-JS-Detection` and a cookie `JS-Detection`, similarly to how the AB testing values are set.

mobaig:
Closing for now: https://trello.com/c/ee87DdP0/350-determine-how-many-people-have-don-t-have-javascript#comment-58a2dfb0b53567cacc648184

PR #23, opened 2017-02-07
Use query string to switch between navigation A/B test variants

suzannehamilton:
The changes to education navigation will be tested by remote testers. Unlike regular users, who should be assigned the A or B variant at random, remote testers need to see a specific version of the A/B test.

Remote testers make their first visit to the site through a link that we email to them. So a convenient way to place them in the A or B buckets is to append a query string in that link that we send them, and identify it when assigning buckets.

The query string will be `?ABTest-EducationNavigation=A`, i.e. one that matches the A/B testing cookie. This isn't essential for this test, but it will make it easier for us to commonise this config for future tests.

https://trello.com/c/F2L41y9R/321-roll-out-the-new-navigation-via-a-b-testing-to-remote-testers-before-the-general-public

Dependencies:
- [ ] #17 (once merged, rebase this PR on master)

### DIFF
--- a/vcl_templates/tldredirect.vcl.erb
+++ b/vcl_templates/tldredirect.vcl.erb
@@ -9,6 +9,14 @@ backend dummy {
 }
 
 sub vcl_recv {
+
+  # Force SSL.
+  if (!req.http.Fastly-SSL) {
+    error 801 "Force SSL";
+  } else {
+    error 802 "Redirect GOV.UK";
+  }
+
   return(error);
 }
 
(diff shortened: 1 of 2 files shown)


MatMoore:
Not really related to the PR, but I was wondering if there was a way in VCL to dynamically set env vars/cookies, so that you could have a common way of setting the env var for all tests, instead of a block for each one?

suzannehamilton:
@MatMoore Yes, it's something @tijmenb and I have discussed. It isn't supported by plain VCL but this is an erb file, so we could generate the VCL config for each test from the erb template and a yaml file. When a new team wants to add an A/B test, they could just add the name, A/B percentages, etc. to the yaml.

The only reason we haven't done this is because of the risk of making such a sweeping change to the CDN config without tests. I think Tijmen's planning to speak to Infrastructure about adding tests, extracting the config (and making some of it public - there's no need for the A/B test config to be in the private repo ), but we haven't had the time to devote to it yet.

suzannehamilton:
Replaced by #29.

PR #24, opened 2017-02-15
[DO NOT MERGE]Replace mirror1 with an S3 bucket

SamLR:
We want to serve our mirrors from an S3 bucket rather than a
dedicated machine. To do this we need to configure fastly with the
correct authorization headers and remove the AWS headers from
returned objects.

This is a placeholder PR until we can fully check this. I believe it to be reasonable correct but a test bed (which isn't production) needs to be made.

### DIFF
--- a/vcl_templates/www.vcl.erb
+++ b/vcl_templates/www.vcl.erb
@@ -68,39 +68,42 @@ backend F_mirror0 {
         .interval = 10s;
     }
 }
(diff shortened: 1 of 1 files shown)


SamLR:
After chatting with @surminus I think this is wrong. The plan is to use an IP based read-only policy on S3 so the authorisation can be removed.

To reflect this I'm going to close this PR and try again.

PR #27, opened 2017-02-22
Add tls cookie

h-lame:
For: https://trello.com/c/JX541tka/113-add-tls-version-information-to-google-analytics

Also see https://github.com/alphagov/static/pull/900 which consumes this cookie.  Our goal here is to set a cookie with the tls protocol version as understood by fastly and then read the cookie in the frontend so we can send the protocol version to GA and let us work out how many visitors use each version of the protocol.  This will ultimately inform our decisions about when/if to drop support older versions of the protocol (see: [Fastly's plan to deprecate 1.0 and 1.1](https://www.fastly.com/blog/phase-two-our-tls-10-and-11-deprecation-plan).

### DIFF
--- a/vcl_templates/www.vcl.erb
+++ b/vcl_templates/www.vcl.erb
@@ -296,16 +296,11 @@ sub vcl_deliver {
     add resp.http.Set-Cookie = "ABTest-Example=" req.http.GOVUK-ABTest-Example "; expires=" now + 1d;
   }
 
-  if (tls.client.protocol == "TLSv1.2") {
-    add resp.http.Set-Cookie = "TLSversion=TLSv1.2; max-age=0; secure";
-  } else if (tls.client.protocol == "TLSv1.1") {
-    add resp.http.Set-Cookie = "TLSversion=TLSv1.1; max-age=0; secure";
-  } else if (tls.client.protocol == "TLSv1") {
-    add resp.http.Set-Cookie = "TLSversion=TLSv1.0; max-age=0; secure";
-  } else {
-    add resp.http.Set-Cookie = "TLSversion=unknown; max-age=0; secure";
+  # Set the tls version session cookie with the raw protocol version from
+  # fastly only if it isn't already set
+  if (req.http.Cookie !~ "TLSversion") {
+    add resp.http.Set-Cookie = "TLSversion=" tls.client.protocol "; secure";
   }
-
 #FASTLY deliver
 }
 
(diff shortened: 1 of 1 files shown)


rubenarakelyan:
https://www.gov.uk/help/cookies will also need to be updated

h-lame:
Thanks @rubenarakelyan - it's in the trello card and we've got a story in zendesk with the content team to sort it out.

vcl_templates/www.vcl.erb
```diff
@@ -296,6 +296,9 @@ sub vcl_deliver {
     add resp.http.Set-Cookie = "ABTest-Example=" req.http.GOVUK-ABTest-Example "; expires=" now + 1d;
   }
 
+  # Set the tls version cookie with the raw protocol version from fastly
+  add resp.http.Set-Cookie = "TLSversion=" tls.client.protocol "; max-age=0; secure";
```
suzannehamilton: I might have just misunderstood this, but wouldn't `max-age=0` mean that the browse never sets the cookie? I thought this tended to be used to delete existing cookies.

suzannehamilton: Also, do you know if there are any repercussions to adding `Set-Cookie` to every single request? I'm not aware of any problems that it would cause - it just seems a bit surprising.

h-lame: re: `max-age=0` I think (and @davidillsley might correct me - I took it from his proof-of-concept) that our intention is to use cookies as a way of getting the value from the CDN into something that can be read by JS in the browser, not necessarily that we want to persist the cookie.  But perhaps we should be using `expires=" now` instead (better cross-browser support).

re: repercussions - I can't think of any other than the increase in payload size which will add up over time, but I have no idea what else we might want to measure or monitor.

timblair: Do we need an explicit expiry, or can we just use a session cookie (by excluding both `max-age` and `expires`)?  We can also check for a pre-existing cookie before setting a new one:

```js
if (req.http.Cookie !~ "TLSversion") {
  add resp.http.Set-Cookie = "TLSversion=" tls.client.protocol "; secure";
}
```

davidillsley: I can't remember why I set `max-age=0`, I didn't test it, and the docs I'm looking at look suggest it's wrong. I was aiming for it being visible to JS, but not stored an sent back to the server.

h-lame: Rebased to include @timblair's idea of having a check to see if it already exists and not setting max-age or expires.

suzannehamilton: LGTM. 👍

h-lame:
Thanks all!

PR #31, opened 2017-03-07
Show the B version of the Education Navigation A/B test to 10% of users

suzannehamilton:
Increase the B fraction from 0/10 to 1/10 to start the Education Navigation A/B test.

Also increase the A/B test cookie expiry from 1 day to 1 year, so that users see the same variant throughout the test.

Marked as DO NOT MERGE because we want to control when we deploy this.

Trello: https://trello.com/c/pCvpvtds/356-roll-out-the-navigation-a-b-test-to-the-general-public

### DIFF
--- a/vcl_templates/www.vcl.erb
+++ b/vcl_templates/www.vcl.erb
@@ -224,11 +224,7 @@ sub vcl_recv {
     # Set the value of the header to whatever decision was previously made
     set req.http.GOVUK-ABTest-EducationNavigation = req.http.Cookie:ABTest-EducationNavigation;
   } else {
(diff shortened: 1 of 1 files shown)


alecgibson:
When we decide we want to ramp up the fraction of B/A, does this mean that we're banking on visitors who haven't been to gov.uk since the A/B test started? I assume we're okay with that?

suzannehamilton:
@alecgibson - yes, I don't think there's an easy way to avoid it.

The 1 day expiry means that it won't be a problem for the initial release, but if we increase the fraction later then some users will get stuck in the A version until we roll it out to everyone.

tijmenb:
Let's go!

PR #35, opened 2017-03-09
Add users to B group when going to a nav page

carvil:
We are enabling the new navigation pages for everyone that goes directly
to them. The reason for this is to avoid people seeing 404s, and search
indexers index pages that might point people to 404s.

This is a complementary change where we assign users to the B group when
they go directly to a new navigation page, so they see a consistent
navigation throughout the site.

All new navigation pages start with `/education`.

cc/ @suzannehamilton @tijmenb

### DIFF
--- a/vcl_templates/www.vcl.erb
+++ b/vcl_templates/www.vcl.erb
@@ -220,6 +220,10 @@ sub vcl_recv {
     set req.http.GOVUK-ABTest-EducationNavigation = "A";
   } else if (req.url ~ "[\?\&]ABTest-EducationNavigation=B(&|$)") {
     set req.http.GOVUK-ABTest-EducationNavigation = "B";
+  } else if (req.url ~ "^/education(/|$)") {
+    # If a user goes to a new navigation page, add them automatically to the B
+    # group, so they can see a consistent navigation throughout.
+    set req.http.GOVUK-ABTest-EducationNavigation = "B";
   } else if (req.http.Cookie ~ "ABTest-EducationNavigation") {
     # Set the value of the header to whatever decision was previously made
     set req.http.GOVUK-ABTest-EducationNavigation = req.http.Cookie:ABTest-EducationNavigation;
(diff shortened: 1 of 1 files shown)


vcl_templates/www.vcl.erb
```diff
@@ -220,6 +220,10 @@ sub vcl_recv {
     set req.http.GOVUK-ABTest-EducationNavigation = "A";
   } else if (req.url ~ "[\?\&]ABTest-EducationNavigation=B(&|$)") {
     set req.http.GOVUK-ABTest-EducationNavigation = "B";
+  } else if (req.url ~ "^\/education(\/)?.*$") {
```
suzannehamilton: I think we need an equivalent check in the delivery step to make sure the B cookie gets set. Otherwise this will show the user the B version on this request but will preserve their original cookie.

carvil: That's a very good point, I've now added it below.

suzannehamilton: Could the regex be simplified to `^/education/` in both places? It's doing a similar thing to line 310, and we don't bother escaping the forward slashes or matching the full path there.

carvil: Yup, I can simplify to `^/education` without the slash so it works on the root taxon page.

suzannehamilton: Are there any content pages at URLs like www.gov.uk/education-and-some-other-info which would be affected by that?

carvil: I am not sure, but it does sound like a possibility. What do you think of rolling back to the more verbose regex: `"^\/education(\/)?.*$"`? That way we guarantee that we set the cookie only for:
- `/education`
- `/education/`
- `/education/<any-taxon-base-path>`

And we know there are no content pages nested under `/education`.

suzannehamilton: Wouldn't that also match `/education-other-stuff`?

What about `^/education(/|$)`?

I'm starting to wonder if this is worth a Smokey test, given that we can't test it any other way.

carvil: I think `^/education(/|$)` is fine. I will add a smokey test for this.

suzannehamilton: Looks like it's still `^/education` ? Or have you just not pushed this change yet?

carvil: I haven't pushed yet, sorry

PR #36, opened 2017-03-10
Allow query params on /education URLs

carvil:
When adding smokey tests to verify we were assigning users to the
expected bucket, we realised we used a cache busting param to the
request. In doing so, the request wasn't recognised as being a request
to /education, meaning we weren't being assigned to the correct bucket.

This change makes sure we also allow for `?` after /education.

### DIFF
--- a/vcl_templates/www.vcl.erb
+++ b/vcl_templates/www.vcl.erb
@@ -220,7 +220,7 @@ sub vcl_recv {
     set req.http.GOVUK-ABTest-EducationNavigation = "A";
   } else if (req.url ~ "[\?\&]ABTest-EducationNavigation=B(&|$)") {
     set req.http.GOVUK-ABTest-EducationNavigation = "B";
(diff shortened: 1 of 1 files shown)


vcl_templates/www.vcl.erb
```diff
@@ -220,7 +220,7 @@ sub vcl_recv {
     set req.http.GOVUK-ABTest-EducationNavigation = "A";
   } else if (req.url ~ "[\?\&]ABTest-EducationNavigation=B(&|$)") {
     set req.http.GOVUK-ABTest-EducationNavigation = "B";
-  } else if (req.url ~ "^/education(/|$)") {
+  } else if (req.url ~ "^/education(\/|\?|\$)") {
```
suzannehamilton: I don't think `\$` should have the backslash - we're matching to the end of the string rather than a literal `$`.

carvil: Thanks, I've removed it now.

PR #37, opened 2017-03-13
Enable www S3 mirror

afda16:
Update www template to use new S3 mirror instead of mirror1.
There are a few changes specific for S3:
- Depending on how the bucket is configured, the backend probe request
needs to point to the directory where the content is located
- S3 does not set cache headers by default, we need to override the default TTL and add
header based on previous Nginx mirror configuration
- Requests need to be rewritten to match location in the bucket and add file extension
if it does not exist in the request, similar to try_files configuration in Nginx
- S3 errors need to serve static error page

### DIFF
--- a/vcl_templates/www.vcl.erb
+++ b/vcl_templates/www.vcl.erb
@@ -68,12 +68,12 @@ backend F_mirror0 {
         .interval = 10s;
     }
 }
-# Mirror backend for provider 1
-backend F_mirror1 {
+# Mirror backend for S3
+backend F_mirrorS3 {
     .connect_timeout = 1s;
     .dynamic = true;
-    .port = "<%= config.fetch('provider1_mirror_port', 443) %>";
-    .host = "<%= config.fetch('provider1_mirror_hostname') %>";
+    .port = "<%= config.fetch('s3_mirror_port', 443) %>";
+    .host = "<%= config.fetch('s3_mirror_hostname') %>";
     .first_byte_timeout = 15s;
     .max_connections = 200;
     .between_bytes_timeout = 10s;
@@ -85,13 +85,13 @@ backend F_mirror1 {
     <%- if config['ssl_ciphers'] -%>
     .ssl_ciphers = "<%= config['ssl_ciphers'] -%>";
     <%- end -%>
-    .ssl_cert_hostname = "<%= config.fetch('ssl_cert_hostname', config.fetch('provider1_mirror_hostname')) %>";
-    .ssl_sni_hostname = "<%= config.fetch('ssl_sni_hostname', config.fetch('provider1_mirror_hostname')) %>";
+    .ssl_cert_hostname = "<%= config.fetch('ssl_cert_hostname', config.fetch('s3_mirror_hostname')) %>";
+    .ssl_sni_hostname = "<%= config.fetch('ssl_sni_hostname', config.fetch('s3_mirror_hostname')) %>";
 
     .probe = {
         .request =
-            "HEAD / HTTP/1.1"
-            "Host: <%= config.fetch('provider1_mirror_hostname') %>"
+            "HEAD <%= config.fetch('s3_mirror_probe_request', '/') %> HTTP/1.1"
+            "Host: <%= config.fetch('s3_mirror_hostname') %>"
             "User-Agent: Fastly healthcheck (git version: <%= config['git_version'] %>)"
             "Connection: close";
         .threshold = 1;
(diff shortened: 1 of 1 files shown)


surminus:
I think this all looks OK, although the regex makes my head hurt a bit 🙂 

I'd like to test it as VCL syntax, how did you go about testing it?

afda16:
I'll add a Varnish VCL to the original ticket so you can test

vcl_templates/www.vcl.erb
```diff
@@ -184,9 +184,20 @@ sub vcl_recv {
...
-    set req.backend = F_mirror1;
-    set req.http.host = "<%= config.fetch('provider1_mirror_hostname') %>";
-    set req.http.Fastly-Backend-Name = "mirror1";
+    set req.backend = F_mirrorS3;
+    set req.http.host = "<%= config.fetch('s3_mirror_hostname') %>";
+    set req.http.Fastly-Backend-Name = "mirrorS3";
+
+    if (req.url ~ "^(/.*)?/([^#?\s]+)\.(atom|css|csv|docx|gif|html|ico|ics|jpeg|jpg|js|json|ods|odt|pdf|png)([\?#]+.*)?$") {
+      # Requests with document extension, rewrite adding bucket directory prefix
+      set req.url = regsub(req.url, "^(.*)$", "/<%= config.fetch('s3_mirror_prefix') %>\1");
```
SamLR: Is there a reason for the regsub over e.g. `set req.url = ""/<%= config.fetch('s3_mirror_prefix') %>" + req.url` ?

vcl_templates/www.vcl.erb
```diff
@@ -184,9 +184,20 @@ sub vcl_recv {
...
   # FIXME: Prefer a fallback director if we move to Varnish 3
   if (req.restarts > 2) {
-    set req.backend = F_mirror1;
-    set req.http.host = "<%= config.fetch('provider1_mirror_hostname') %>";
-    set req.http.Fastly-Backend-Name = "mirror1";
+    set req.backend = F_mirrorS3;
+    set req.http.host = "<%= config.fetch('s3_mirror_hostname') %>";
+    set req.http.Fastly-Backend-Name = "mirrorS3";
+
+    if (req.url ~ "^(/.*)?/([^#?\s]+)\.(atom|css|csv|docx|gif|html|ico|ics|jpeg|jpg|js|json|ods|odt|pdf|png)([\?#]+.*)?$") {
```
SamLR: This feels like it could be clearer. As I understand it it's testing:

1. Is this a request for something with an explicit extension? If so add the s3-prefix to the request URL
2. Is this a request for the root. If so add the s3-prefix and 'index.html'
3. Is this a request for something that doesn't have an extension, if so add '.html'

Would be clearer as:

1. is this a request for root: make the url `index.html`
2. does this have an extension: if not add '.html'
3. add the prefix

e.g. 
```
if (req.url ~ "^/?$"){
  set req.url = "/index.html";
} else if (  req.url !~ "\." ){ # i.e. the URL doesn't contain a '.'
  set req.url = "/" + req.url + ".html";
}
set req.url = "<%= config.fetch('s3_mirror_prefix') %>" + req.url
```

SamLR: This may be wrong but regardless I'm worried about the explicit list of extensions in line 191 which I suspect will be fragile

SamLR: This feels like it could be clearer. As I understand it it's testing:

1. Is this a request for something with an explicit extension? If so add the s3-prefix to the request URL
2. Is this a request for the root. If so add the s3-prefix and 'index.html'
3. Is this a request for something that doesn't have an extension, if so add '.html'

Would be clearer as:

1. is this a request for root: make the url `index.html`
2. does this have an extension: if not add '.html'
3. add the prefix

e.g. 
```
if (req.url ~ "^/?$"){
  set req.url = "/index.html";
} else if (  req.url !~ "\." ){ # i.e. the URL doesn't contain a '.'
  set req.url = "/" + req.url + ".html";
}
set req.url = "<%= config.fetch('s3_mirror_prefix') %>" + req.url
```

afda16: We have URLs that contain a dot in the filename, so we need to differentiate when it's an extension and when it's not. The list of extensions will need to be reviewed, and I agree it's fragile, but unless we force people not to publish filenames that contain a dot I don't know what else we can do

The s3 prefix needs to be added to all the requests, that is the path where the files are stored in the S3 bucket

surminus:
I think a bunch of commits have been accidentally included.

afda16:
I remove the commits and tidy up the expressions, with extensions up-to-date

surminus:
I think this looks fine :+1:

PR #39, opened 2017-03-16
Js detection cookie

carolinegreen:
Opening #22 again. This is a re-run of the experiment described in https://gds.blog.gov.uk/2013/10/21/how-many-people-are-missing-out-on-javascript-enhancement/

The original experiment didn't use cookies, but we're using cookies as an enhancement to track unique users and possibly filter out any pre-scanning behaviour by browsers which could possibly skew the number of users appearing to use javascript.

https://www.gov.uk/help/cookies now has a description of this cookie.


carolinegreen:
@alexmuller I'm pairing with @afda16 on these requests and trying to get them through.  She said that you have more context on this and would like you to check that this request is okay to go now if that's okay?

PR #41, opened 2017-03-16
Enable assets S3 mirror

afda16:
Update assets template to use new S3 mirror instead of mirror1.
There are a few changes specific for S3:
- Depending on how the bucket is configured, the backend probe request
needs to point to the directory where the content is located
- S3 does not set cache headers by default, we need to override the default TTL and add
header based on previous Nginx mirror configuration
- Requests need to be rewritten to match location in the bucket
- S3 errors need to serve static error page


surminus:
:+1: still looks good

PR #42, opened 2017-03-21
Add cookie for GOVUK-ABTest-Benchmarking

carvil:
### Background

We are going to be running a new A/B test for the GOV.UK benchmarking
task. In this A/B test, which should last less than 1 week, we will be
adding a tracking library so we understand how users interact with the
new navigation pages we have exposed for Education content.

### Changes

In this commit we add rules to set the ABTest-Benchmarking cookie based
on a few things.

On `vcl_recv`:

- If the user visits GOV.UK with a parameter like
  `?ABTest-Benchmarking=B`, we will use that value to set the header
  accordingly, so the application knows what variant to render (and set
  the `Vary` header appropriately);
- If the user already has a cookie set, we use that value in the
  header, so the application knows what variant to render (and set
  the `Vary` header appropriately);
- Finally, for anyone else, we always put them in the `A` bucket,
  because we only want to track users in the benchmarking test and no
  one else.

On `vcl_deliver`, on the first visit, we set the cookie and make sure
it expires after 7 days.

Trello: https://trello.com/c/NJ6O76BR/281-heatmap-tracking

### DIFF
--- a/vcl_templates/www.vcl.erb
+++ b/vcl_templates/www.vcl.erb
@@ -235,7 +235,7 @@ sub vcl_recv {
       set req.http.GOVUK-ABTest-EducationNavigation = "A";
     }
   }
-  
+
   if (req.http.Cookie ~ "JS-Detection") {
     set req.http.GOVUK-JS-Detection = req.http.Cookie:JS-Detection;
   } else {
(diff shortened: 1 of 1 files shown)


vcl_templates/www.vcl.erb
```diff
@@ -235,6 +235,21 @@ sub vcl_recv {
...
+  } else if (req.url ~ "[\?\&]ABTest-Benchmarking=B(&|$)") {
+    set req.http.GOVUK-ABTest-Benchmarking = "B";
+  } else if (req.http.Cookie ~ "ABTest-Benchmarking") {
+    # Set the value of the header to whatever decision was previously made
+    set req.http.GOVUK-ABTest-Benchmarking = req.http.Cookie:ABTest-Benchmarking;
+  } else {
+    # We default everyone to A, because remote testers will have a param in the
+    # URL to add them to the B group. In this test, we don't want anyone else
+    # having the B variant.
+    set req.http.GOVUK-ABTest-Benchmarking = "A";
```
carvil: I'm tempted to remove this last `else` and not assign any cookie/header to non-testers. This way, we don't `pollute` the cookies of most GOV.UK users. Thoughts?

carvil:
Thanks @suzannehamilton . I've sent an email so the cookies page gets updated. I will add a trello card so we chat about naming of things related to A/B tests.

Issue #14, opened 2017-01-13
Adding HSTS to the raw gov.uk domain


davidillsley:
We've got HSTS set up for www.gov.uk, but not for the raw domain. I assume this is because we were paranoid about accidental application of `includeSubdomains` (we don't want to disable access to other uses of gov.uk which aren't yet https enabled).

We should do better. I'd propose the following:
- [x] Change http://gov.uk to redirect to https://gov.uk rather than https://www.gov.uk
- [x] Add an HSTS header to https://gov.uk responses (using VCL) *without* `includeSubdomains` and with a very short `max-expiry` (eg 10 seconds)
- [x] Increase the `max-expiry` in a couple of increments to decrease the impact of an error
- [ ] Request manual preload into Chrome (automatic preload wont work because of the lack of `includeSubdomains`)

Sound sensible? PRs doing that accepted?

alexmuller:
Pull requests very much accepted! We have a story for this in our backlog but if you get there first that would be wonderful.

issyl0:
The three people who worked on this don't work here anymore, but it looks from the PRs like the work has been done. Closing.

PR #45, opened 2017-03-29
Bump TLD redirect HSTS max-age to 2 years

afda16:
The 1 hour max-age introduced in #26 seems to be fine. This
commit increases that to 2 years (63072000 seconds), as a further step
outlined in #14.


SamLR:
This looks like the correct number so go for it

PR #46, opened 2017-04-04
Serve an empty response direct from Fastly

boffbowsh:
This will be used to store logs of the requests from the tracking
Javascript.

### DIFF
--- a/vcl_templates/assets.vcl.erb
+++ b/vcl_templates/assets.vcl.erb
@@ -153,6 +153,11 @@ sub vcl_recv {
      error 801 "Force SSL";
   }
 
+  # Serve and log an empty response for analytics purposes
+  if (req.url ~ "^/static/a\?") {
+     error 802 "Empty for analytics logging";
+  }
+
   # Serve from stale for 24 hours if origin is sick
   set req.grace = 24h;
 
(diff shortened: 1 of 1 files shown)


vcl_templates/assets.vcl.erb
```diff
@@ -153,6 +153,11 @@ sub vcl_recv {
      error 801 "Force SSL";
   }
 
+  # Serve and log an empty response for analytics purposes
+  if (req.url == "/static/a") {
```
afda16: does the URL receive parameters? I think req.url might include them, maybe ~ is safer

boffbowsh: Yes, good catch, thanks! Amended.

afda16:
It looks good to me

boffbowsh:
Thanks! Can you merge it and deploy it to staging for testing please?

vcl_templates/assets.vcl.erb
```diff
@@ -153,6 +153,11 @@ sub vcl_recv {
      error 801 "Force SSL";
   }
 
+  # Serve and log an empty response for analytics purposes
+  if (req.url ~ "^/static/a") {
```
SamLR: Won't this capture any url beginning with 'static/a' as it's unterminated? Is it possible to scope this more tightly?

boffbowsh: Also a good point, and fixed. In reality it will only be called with a querystring, so it now must begin with `/static/a?`. 

If someone calls it with `/static/a`, it will fall through to the backend and 404.

PR #49, opened 2017-04-27
Add Fastly-Backend-Name header on all Assets responses

afda16:
Match www and assets configuration by adding the Fastly-Backend-Name
header to Assets responses. This header allows us to monitor which
backend is serving each response.


surminus:
Needs a rebase.

afda16:
Rebased

PR #51, opened 2017-05-04
Add search match length a/b test config

brenetic:
This will be used for the minimum should match a/b test in order to measure improvements to search.

https://trello.com/c/43eBQfwf/80-prepare-the-infrastructure-for-the-minimum-should-match-a-b-test

### DIFF
--- a/vcl_templates/www.vcl.erb
+++ b/vcl_templates/www.vcl.erb
@@ -250,7 +250,28 @@ sub vcl_recv {
       set req.http.GOVUK-ABTest-EducationNavigation = "A";
     }
   }
-  
+
+  <% ['SearchMatchLength'].each do |test_name| %>
+  if (req.http.User-Agent ~ "^GOV\.UK Crawler Worker") {
+    set req.http.GOVUK-ABTest-<%= test_name %> = "A";
+  # Some users, such as remote testers, will be given a URL with a query string
+  # to place them into a specific bucket.
+  } else if (req.url ~ "[\?\&]ABTest-<%= test_name %>=A(&|$)") {
+    set req.http.GOVUK-ABTest-<%= test_name %> = "A";
+  } else if (req.url ~ "[\?\&]ABTest-<%= test_name %>=B(&|$)") {
+    set req.http.GOVUK-ABTest-<%= test_name %> = "B";
+  } else if (req.http.Cookie ~ "ABTest-<%= test_name %>") {
+    # Set the value of the header to whatever decision was previously made
+    set req.http.GOVUK-ABTest-<%= test_name %> = req.http.Cookie:ABTest-<%= test_name %>;
+  } else {
+    if (randombool(3,10)) {
+      set req.http.GOVUK-ABTest-<%= test_name %> = "B";
+    } else {
+      set req.http.GOVUK-ABTest-<%= test_name %> = "A";
+    }
+  }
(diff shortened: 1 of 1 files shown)


vcl_templates/www.vcl.erb
```diff
@@ -250,7 +250,28 @@ sub vcl_recv {
       set req.http.GOVUK-ABTest-EducationNavigation = "A";
     }
   }
-  
+
+  <% ['SearchMatchLength'].each do |test_name| %>
+  if (req.http.User-Agent ~ "^GOV\.UK Crawler Worker") {
```
suzannehamilton: This is just the `vcl_recv` half of the CDN config, which adds a header to the request that's sent to the origin service. To send the cookie to the user, we'll have to update `vcl_deliver` as well. Are we intentionally skipping that step while we implement the rummager logic?

brenetic: Updated the PR

suzannehamilton: Do you think we can skip the crawler logic? I don't think the crawler is going to see search results, so we can probably drop this (unless we think this logic is going to be shared by the performance platform, who might care about which version the crawler saves).

sihugh: If this is intended for reuse (we'd think we might like to use it), then leaving the crawler worker stuff in would be good for us.

vcl_templates/www.vcl.erb
```diff
@@ -354,7 +375,11 @@ sub vcl_deliver {
   if (req.http.Cookie !~ "ABTest-EducationNavigation" || req.url ~ "[\?\&]ABTest-EducationNavigation=" || req.url ~ "^/education(\/|\?|$)") {
     add resp.http.Set-Cookie = "ABTest-EducationNavigation=" req.http.GOVUK-ABTest-EducationNavigation "; expires=" now + 365d "; path=/";
   }
-
+  <% ['SearchMatchLength'].each do |test_name| %>
+  if (req.http.Set-Cookie !~ "ABTest-<%= test_name %>" || req.url ~ "[\?\&]ABTest-<%= test_name %>=") {
+    add resp.http.Set-Cookie = "ABTest-<%= test_name %>=" req.http.GOVUK-ABTest-<%= test_name %> "; expires=" now + 1d "; path=/";
```
dwhenry: Any reason this has an expiry of 1day v's the education nav which used 365days?

brenetic: 365 just seemed excessive. I was also thinking that searches might happen multiple times with different terms each time.

suzannehamilton: Yes, the education navigation test has a long expiry time so that users see a consistent version of the site (always A or always B). But it means that there's a lag when we change the B percentage, because some users are stuck in A.

We set the search A/B test expiry to 1 day because thought it would be OK for users to see slightly different search results from day to day, and it gives us more flexibility to change the B percentage. It'd be good to know what Tara and Mark think about that, if we haven't checked with them already?

dwhenry: 👍

dwhenry: 👍

brenetic:
Close this in favour of https://github.com/alphagov/govuk-cdn-config/pull/52

PR #52, opened 2017-05-12
Add extra templating for ab testing

brenetic:
This adds dynamic template rendering for both the setting of the request headers used for A/B tests and for the setting of the cookies.

Configuration for A/B tests is now done in [`fastly-configure`](https://github.com/alphagov/fastly-configure). Example config can be seen in [this PR](https://github.com/alphagov/fastly-configure/pull/21)

Using Fastly dictionaries was investigated to some length and we got it working, however it would have required a much larger vlc template with more generalised test names (test_0, test_1, etc)

https://trello.com/c/jsmd99NV/89-prototype-a-b-testing-configuration-changer

### DIFF
--- a/vcl_templates/www.vcl.erb
+++ b/vcl_templates/www.vcl.erb
@@ -250,7 +250,36 @@ sub vcl_recv {
       set req.http.GOVUK-ABTest-EducationNavigation = "A";
     }
   }
-  
+
+  # Begin dynamic section
+  <% ab_tests.each do |test| %>
+  if (req.http.User-Agent ~ "^GOV\.UK Crawler Worker") {
+    set req.http.GOVUK-ABTest-<%= test['name'] %> = "A";
+
+  } else if (req.url ~ "[\?\&]ABTest-<%= test['name'] %>=A(&|$)") {
+    # Some users, such as remote testers, will be given a URL with a query string
+    # to place them into a specific bucket.
+    set req.http.GOVUK-ABTest-<%= test['name'] %> = "A";
+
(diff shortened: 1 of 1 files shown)


vcl_templates/www.vcl.erb
```diff
@@ -250,7 +250,41 @@ sub vcl_recv {
       set req.http.GOVUK-ABTest-EducationNavigation = "A";
     }
   }
-  
+
+  <% 5.times do |i| %>
```
SamLR: why 5?

brenetic: This is an arbitrary number chosen to represent five 'slots' which teams can use for ab testing. This can be larger or smaller depending on what the teams feel is required.

SamLR:
I think that looks reasonable.

vcl_templates/www.vcl.erb
```diff
@@ -355,6 +384,13 @@ sub vcl_deliver {
     add resp.http.Set-Cookie = "ABTest-EducationNavigation=" req.http.GOVUK-ABTest-EducationNavigation "; expires=" now + 365d "; path=/";
   }
 
+  # Begin dynamic section
+  <% ab_tests.each do |test| %>
+  if (req.http.Cookie !~ "ABTest-<%= test['name'] %>" || req.url ~ "[\?\&]ABTest-<%= test['name'] %>") {
```
suzannehamilton: We should probably add the crawler test to this condition - see line 378. I don't think it will make any practical difference, but it seems wrong add `Set-Cookie` to every response to the crawler.

PR #58, opened 2017-06-30
Guard against no running a/b tests

brenetic:
We now check for any existing a/b tests in the config yaml file.


brenetic:
cc @rubenarakelyan

PR #61, opened 2017-10-25
Add dynamic multivariate vcl templating

steventux:
https://trello.com/c/gsUWTMyL/34-set-up-infrastructure-for-multi-variant-tests-multi-variant

Adds the ability to define multivariate test configurations with allocations based on percentages or arbitrary nominators, the denominator is the sum of nominators.
Requires that variants are specified in [alphagov/fastly-configure](https://github.com/alphagov/fastly-configure/commit/578518f74f9e8b9861fe4c1ccfdf3b23978d2c23).
The allocations are [stored in Fastly dictionaries](https://github.com/alphagov/cdn-configs/commit/a08784f92c718c86c2b2a2a0fb4d3a0208b71aa1) so they can be adjusted without a redeployment of the Fastly VCL configuration, this means we can still switch off tests by allocating 100% users to the first variant.
Each test will require its [own dictionary containing the allocations for each variant](https://github.com/alphagov/cdn-configs/blob/a08784f92c718c86c2b2a2a0fb4d3a0208b71aa1/fastly/dictionaries/config/multivariatetest_percentages.yaml).

I ran a test of this branch on staging deploying the test dictionaries and CDN then took a 1000 request sample of header allocations:

```
for i in `seq 1 1000`;
do
  curl -Ik https://www.staging.publishing.service.gov.uk/ | grep Multivariate | awk '/=*.?;/ {print $2}' | cat >> mv.txt
  sleep 1
done

$ grep firstVariant mv.txt | wc -l
502
$ grep secondVariant mv.txt | wc -l
250
$ grep thirdVariant mv.txt | wc -l
248
```

Depends on some dictionary config for existing tests https://github.com/alphagov/cdn-configs/pull/6 and https://github.com/alphagov/fastly-configure/pull/42

cc @tijmenb @fofr @suzannehamilton


suzannehamilton:
This is awesome. 💃 

By moving the bucket config into this repo, does it mean that we'll have to do a full Fastly deploy to change the percentage of users in each bucket? It would be good to avoid this if possible - it's less risky to just deploy the dictionaries.

@tijmenb suggested keeping the bucket config in the dictionaries repo and just specifying the overall percentage of users who are in a test bucket. It would give you less flexibility because you would have to split the users evenly between B/C/D/etc., but it might be worth it to keep the deployment simple?

tijmenb:
👍 Looking great.

To add to @suzannehamilton's comments, when Carlos first did the multivariate work they made it so that they could use descriptive names instead of "A" and "B", so the code in the app reads like `variant?('ShowBlueBox')` instead of `variant?("C")`. Could we support that?

https://github.com/alphagov/collections/blob/eed9aa5db0939eb30179d13d6b0a3199c8600b5d/app/models/blue_box_ab_test_request.rb#L24-L26

Also, would it be worth it to use rspec for consistency with other apps?

steventux:
@suzannehamilton Thanks for reviewing, you're right it's a nasty limitation. It might be possible to achieve this using a dictionary per MV test. I'll give this a try.

steventux:
@suzannehamilton @tijmenb [I've added a WIP commit](https://github.com/alphagov/govuk-cdn-config/pull/61/commits/d592d1800cfc3c0d0dab51a51b9a5ca2edf695fd) exploring the possibility of using a dictionary per MV test. The dictionaries would contain `variant-name : nominator` pairs (eg. `someVariant : 50`), which addresses naming variants rather than assigning a letter. We'd still need to deploy Fastly config to pick up new tests but at least the variant allocation lookups would be dynamic.

This method declares and sets the nominator/denominator vars per test and variant in the vcl code, there's still a reliance on the template-level config knowing how many variants there are so this would look more like:

```
- MyTest:
  - variant1
  - variant2
``` 

The other drawback to this approach is the lack of ability to test the logic now pushed into the vcl, is there a safe way of validating and running this output?
I've not attempted to do anything complex with maths in the vcl code as I don't have much confidence in how it will work, so splitting the test variants equally hasn't been addressed here, I can't think of a way to achieve that with the limited functions and options of vcl.

steventux:
@suzannehamilton @brenetic  would you mind taking another look at this PR. I've deployed this branch a couple of times and run tests on staging the PR description above has updated info.

PR #65, opened 2018-03-05
Add TLS header for Licensify requests

PeteLeaman:
Added a request header containing the TLS version number for requests to the Licensify app
(This is so we can block requests with TLS versions below 1.2 for payment security reasons)

### DIFF
--- a/vcl_templates/www.vcl.erb
+++ b/vcl_templates/www.vcl.erb
@@ -207,6 +207,11 @@ sub vcl_recv {
   # Unspoofable original client address.
   set req.http.True-Client-IP = req.http.Fastly-Client-IP;
 
+  # Set TLS version request header for requests to the Licensify application
+  if (req.url ~ "^/apply-for-a-licence/.*") {
+    set req.http.TLSversion = tls.client.protocol;
+  }
+
 #FASTLY recv
 
   if (req.request != "HEAD" && req.request != "GET" && req.request != "FASTLYPURGE") {
--- a/vcl_templates/www.vcl.erb
+++ b/vcl_templates/www.vcl.erb
@@ -207,7 +207,8 @@ sub vcl_recv {
   # Unspoofable original client address.
   set req.http.True-Client-IP = req.http.Fastly-Client-IP;
 
-  # Set TLS version request header for requests to the Licensify application
+  # Set a TLSversion request header for requests going to the Licensify application
+  # This is used to block unsecure requests at the application level for payment security reasons and an absence of caching in Licensify
   if (req.url ~ "^/apply-for-a-licence/.*") {
     set req.http.TLSversion = tls.client.protocol;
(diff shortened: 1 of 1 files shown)


PeteLeaman:
@MatMoore I think we're not blocking all requests under the min TLS because we'll still let them complete a licence application if that licence doesn't require a fee payment, as some of them don't. So we're going to decide to block them or not on the app itself
Also I think Licensify doesn't use caching, at least I was told

MatMoore:
Sorry, I meant cached through the CDN. Basically this logic at the bottom of `vcl_recv` decides whether to hit the app or not

```
  if (req.request != "HEAD" && req.request != "GET" && req.request != "FASTLYPURGE") {
    return(pass);
  }

[...]

  return(lookup);
```

So POST requests always go straight to the app, but GET requests get looked up first.

I think that's still problematic - I'm envisioning a scenario where the user fills out an application and then gets blocked only when they submit it.

PeteLeaman:
@MatMoore Ah, I see what you mean.

I think @timblair thought about this and decided the best option is to set a header at the moment. I can't remember but think that Licensify is one of the few apps that doesn't use the CDN cache?

Also we're still deciding from our content designer whether to let them try to submit and show an error, or redirect them as soon as they reach the page.

MatMoore:
Is this served on a service domain, or www.gov.uk? Maybe Tim meant requests don't hit the CDN at all (in which case the PR won't have any effect). I think the CDN is only in front of www, but someone who knows the infrastructure better will be able to confirm that.

If requests don't actually hit the CDN, I think you'll need to set the header in a load balancer instead.

If you can find load balancer logs for your app in kibana you can see where the traffic's coming from since requests coming via the CDN will all have the same source IP.

timblair:
All paths beginning with `/apply-for-a-licence` are proxied through to Licensify. _Almost_ all paths are uncached (the response includes `Cache-Control: no-cache, no-store` as a header). Only requests to "Step 1 of ..." pages [like this](https://www.gov.uk/apply-for-a-licence/temporary-event-notice/bracknell-forest/apply-1) are cached (`Cache-Control: max-age=900`).

GA shows fewer than 10,000 views per week across all pages matching `^/apply-for-a-licence/.*/.*/apply-[0-9]+$`. With only a 15 minute cache timeout, the assumption is that the majority of these requests are already hitting the origin, so we're going to just set `Cache-Control: no-cache, no-store` on that too.

Once we've done that, we can pick up the TLS version at both the form download and form submission points (neither of which will be cached), redirecting the user to contact the relevant authority to process their application offline.

PeteLeaman:
Great, I'll add the code to prevent caching the Step 1 page in the tls-blocking branch of Licensify.

In terms of this PR, are we good to go ahead with it as we'll be needing the TLS header?

PR #89, opened 2019-01-25
Merge alphagov/fastly-configure into this repo

tijmenb:
The CDN configuration for GOV.UK is in a bit of a sorry state, caused by many iterations of how we use our CDN (changing providers, adding A/B tests, changes in the deployment process).

We've ended up with 3 repos that are [combined at deploy time](https://github.com/alphagov/fastly-configure/blob/5ef705da6c559cdfa7b740f4bd9454e2143f239b/jenkins.sh) to generate the Varnish configuration (VCL) that we send to Fastly.

- This repo (https://github.com/alphagov/govuk-cdn-config), which contains the VCL templates, but also code to combine generate the templates and [check that they're correct](https://github.com/alphagov/govuk-cdn-config/blob/d2e9210354ba93238f9cb33d75af7e53aa9eee10/spec/www_vcl_erb_spec.rb).
- https://github.com/alphagov/fastly-configure contains code to send the VCL to Fastly, but has also grown to include GOV.UK-specific things like the [deploy script](https://github.com/alphagov/fastly-configure/blob/5ef705da6c559cdfa7b740f4bd9454e2143f239b/jenkins.sh) and [A/B testing configuration](https://github.com/alphagov/fastly-configure/tree/5ef705da6c559cdfa7b740f4bd9454e2143f239b/ab_tests).
- https://github.com/alphagov/cdn-configs (#privaterepo) contains configuration that is passed into the VCL templates. It should only contain secrets, but contains a lot of public information.

We've [documented this as tech debt](https://trello.com/c/y6MIgxjp) on the Trello board. 

This PR merges the https://github.com/alphagov/fastly-configure repo into this one. This will make it easier to understand what is going on, will allow us to end-to-end test the VCL generation, and will 
allow us to start refactoring the system more.

I've used the following to merge in the complete history of the fastly-configure repo:

```
git remote add -f fastly-configure-repo git@github.com:alphagov/fastly-configure.git
git checkout -b merge-in-fastly-configure
git merge --allow-unrelated-histories fastly-configure-repo/master
```

After that, I've solved some merge conflicts and made the deploy script work with the new setup.

After this is merged: 

- [ ] update the Jenkins task to run the `jenkins.sh` in this repo (https://github.com/alphagov/govuk-puppet/pull/8586)
- [ ] update developer docs (https://github.com/alphagov/govuk-developer-docs/pull/1482)
- [ ] archive alphagov/fastly-configure


deploy_vcl
```diff
@@ -0,0 +1,127 @@
...
+  })
+  settings.save!
+end
+
+def upload_vcl(version, contents)
+  vcl_name = 'main'
+
+  begin
+    version.vcl(vcl_name) && version.delete_vcl(vcl_name)
+  rescue Fastly::Error
```
bevanloon: I realise it was like this in https://github.com/alphagov/fastly-configure, but is it not useful to puts the error just for visibility? I don't really know how useful that would be or how often it even occurs.

tijmenb:
I've tested this branch on integration and it works, so merging away: 

https://deploy.integration.publishing.service.gov.uk/job/Deploy_CDN/21/console

deploy_vcl
tijmenb: Good spot, I'll definitely look into that.

PR #105, opened 2019-02-12
Remove Related Links A/A test

karlbaker02:
This PR removes the related links A/A test. We have the required data for analysis of our measurement metrics and as a result no longer need to run this test.


sihugh:
Need to resolve conflicts, but we could then merge

PR #106, opened 2019-02-12
Cache responses to POSTs that are the result of a client error

timblair:
As a general precaution (and to somewhat protect the origin servers) cache the responses to POST requests that are the result of a client error (e.g. a POST that results in a 404).

Also fix up the synthetic response to forced-404s for `autodiscover.xml` paths, and use the same method to quick-fail for non-supported HTTP methods.

### DIFF
--- a/spec/test-outputs/www-integration.out.vcl
+++ b/spec/test-outputs/www-integration.out.vcl
@@ -217,11 +217,8 @@ sub vcl_recv {
 
 #FASTLY recv
 
-  if (req.request != "HEAD" && req.request != "GET" && req.request != "FASTLYPURGE") {
-    return(pass);
-  }
-
-    # Begin dynamic section
+  if (req.request != "POST") {
+        # Begin dynamic section
 if (table.lookup(active_ab_tests, "Example") == "true") {
   if (req.http.User-Agent ~ "^GOV\.UK Crawler Worker") {
     set req.http.GOVUK-ABTest-Example = "A";
@@ -352,6 +349,7 @@ if (table.lookup(active_ab_tests, "RelatedLinksABTest1") == "true") {
 }
 # End dynamic section
 
+  }
 
   return(lookup);
 }
@@ -385,6 +383,12 @@ sub vcl_fetch {
   }
 
   if (beresp.http.Cache-Control ~ "no-(store|cache)") {
+    return (pass);
+  }
+
+  # Pass through good responses to POST requests (i.e. cache responses to POST
+  # requests that are the result of client error).
+  if (req.request == "POST" && !(beresp.status >= 400 && beresp.status <= 499)) {
     return (pass);
   }
 
--- a/spec/test-outputs/www-integration.out.vcl
+++ b/spec/test-outputs/www-integration.out.vcl
@@ -148,6 +148,12 @@ sub vcl_recv {
      error 801 "Force SSL";
   }
 
+  # Reject any methods other than the ones we support.
+  # No need to include FASTLYPURGE here as that's handled above.
+  if (req.request != "HEAD" && req.request != "GET" && req.request != "POST") {
+    error 804 "Invalid Method";
+  {
+
   # Serve a 404 Not Found response if request URL matches "/autodiscover/autodiscover.xml"
   if (req.url.path ~ "(?i)/autodiscover/autodiscover.xml$") {
     error 804 "Not Found";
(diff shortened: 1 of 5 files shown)


spec/test-outputs/www-integration.out.vcl
```diff
@@ -148,9 +148,15 @@ sub vcl_recv {
      error 801 "Force SSL";
   }
 
+  # Reject any methods other than the ones we support.
+  # No need to include FASTLYPURGE here as that's handled above.
+  if (req.request != "HEAD" && req.request != "GET" && req.request != "POST") {
+    error 804 "Invalid Method";
+  {
```
tijmenb: This looks odd, should it be `}`?

Would it be better to change this to a [405 Method Not Allowed](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/405.)? I'm thinking that in a couple of years somebody will use a PUT or PATCH in a frontend app and not understand why it 404s on production - a 405 will be more revealing.

spec/test-outputs/www-integration.out.vcl
```diff
@@ -382,6 +386,12 @@ sub vcl_fetch {
     return (pass);
   }
 
+  # Pass through good responses to POST requests (i.e. cache responses to POST
+  # requests that are the result of client error).
+  if (req.request == "POST" && !(beresp.status >= 400 && beresp.status <= 499)) {
```
kevindew: The numbers here seem potentially too broad. It feels like we could legitimately be returning a 400, 403 or a 422 to a user based on their request that we wouldn't want to cache. 

Off the top of my head I think it's only 404 and 405 we'd want to cache

spec/test-outputs/www-integration.out.vcl
```diff
@@ -382,6 +386,12 @@ sub vcl_fetch {
...
+    return (pass);
```
kevindew: Maybe I'm missing something - but I'm not sure what this actually does as a change. As I understand it tells varnish not to cache POST requests that aren't in 4xx status. But not caching is the default behaviour I'm not sure what actually tells this to cache them.

spec/test-outputs/www-integration.out.vcl
timblair: Yes, it should be `}`, that's a typo.

Technically `405` is indeed the right response type. In the past I've had folks doing security testing saying that it's better to be more ambiguous about the return value (e.g. "`404` tells me that there's nothing there" vs. "`405` tells me there might be something here and I should keep digging.") But personally I'm happy to make it a "proper" `405`.

spec/test-outputs/www-integration.out.vcl
timblair: As far as I'm aware, the behaviour without an explicit `return` is to fall through to the overridden function, and the [default behaviour in that](https://github.com/varnish/Varnish-Cache/blob/2.1/bin/varnishd/default.vcl#L114) is `return (deliver)`. But I'm not exactly a VCL / Varnish expert, so my understanding my well be flawed...

spec/test-outputs/www-integration.out.vcl
timblair: That sounds reasonable. I limited it to client errors initially (we don't want to cache things for a server error, for example), but starting with a smaller subset until necessary it's probably sensible.

spec/test-outputs/www-integration.out.vcl
kevindew: > But I'm not exactly a VCL / Varnish expert, so my understanding my well be flawed...

That's my position too, so take my comments with a pinch of salt 😄 

Looking again I see that this change you have https://github.com/alphagov/govuk-cdn-config/pull/106/files#diff-c789eaeb1ab164ae6d3fb552b5b855d7L214 is the bit that I was missing. That previously prevented the cache lookup.

From my frantic googling to understand this a common theme seems to be that by passing a POST to lookup we end up with [varnish sending a GET request to the backend regardless of which method the requester](https://docs.varnish-software.com/tutorials/caching-post-requests/#step-4-make-sure-the-backend-gets-a-post-request) used. I'm not sure if this behaviour is still current but thought worth mentioning.

timblair: That's an interesting guide that I'd not seen before. I need to think about this some more...

timblair:
This needs more thought, so I'll close it for now. I've pulled ab9d6b8 out into a separate PR (#107) as it's not directly relevant, and has separate value.

PR #121, opened 2019-02-14
Changing CDN configuration for asset-origin around

schmie:
- For AWS migration in staging, we switch around the asset handling
routes on the level of Fastly

- The new default is AWS, holding the frontends, which special cases of
the asset-manager and whitehall-frontend still remaining in Carrenza

solo: @schmie


schmie:
This has been superseded

PR #125, opened 2019-02-20
Add logic to re-route asset traffic

schmie:
[DO NOT MERGE]: While this is safe to merge, deployment would break Fastly ATM - so keep it out of the loop for now.

- During the migration the asset-manager and whitehall assets remains in Carrenza
environments close to Publishing-API and Whitehall, whereas frontends
are located in AWS.
- Assets are served both from asset-manager and frontends, so we need to
separate the traffic for Carrenza and AWS
- We have decided to put the logic to do this into Fastly.
- We filter assets paths with regexp
- Carrenza-side this requires creation of a new nginx proxy
assets-carrenza in order to be able to turn off the cache boxes
post-migration

solo: @schmie

### DIFF
--- a/spec/test-outputs/assets-integration.out.vcl
+++ b/spec/test-outputs/assets-integration.out.vcl
@@ -1,4 +1,4 @@
-backend F_origin {
+backend F_carrenzaorigin {
     .connect_timeout = 5s;
     .dynamic = true;
     .port = "443";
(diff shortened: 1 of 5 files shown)


fredericfran-gds:
$ `curl https://assets-origin.staging.govuk.digital/__canary__ -H "Host: assets-origin.staging.govuk.digital"`
{"message": "Tweet tweet"}

PR #137, opened 2019-03-21
Add Related Links test v4

karlbaker02:
This PR adds the related links test v4 as an A/B/C, described at https://www.gov.uk/help/cookies#multivariate-testing. The cookie age is set to 12 hours initially, with all traffic to the B variant, to help put less stress on the system. We will increase the B/C split afterwards.

The A variant will not be actively set by the CDN, but instead measures any users who "drop through" in our A/B testing framework, so that we can exclude these users from our analysis.


karlbaker02:
@sihugh Yes we can 🙂 The reason we haven't so far (apart from the A/A PR that I need to merge) is because we want to wait until the end of the cookie expiry (4 weeks) before removing, but I appreciate that there are probably some tests already at this stage.

This is the last test we anticipate to run for the ML algorithms, so I will clean everything up next week 👍

PR #148, opened 2019-04-10
Restore ability to purge URLs by Fastly web UI

kevindew:
After a discussion with Fastly Support we learnt that we don't have the
ability to purge individual URLs from Fastly via the web UI because of
our VCL configuration. We actually return a 403 as the requests are not
in the allowed list.

Fastly support suggested that we change our return of error 403 to
instead set a head a header that authentication is required. Without a
valid token to authenticate these will deny the request and return a
message that authentication is needed.

### DIFF
--- a/spec/test-outputs/apt-integration.out.vcl
+++ b/spec/test-outputs/apt-integration.out.vcl
@@ -25,9 +25,9 @@ backend F_apt {
 sub vcl_recv {
 #FASTLY recv
 
-    # Prevent third parties from flushing CDN caches with FASTLYPURGE
+    # Require authentication for FASTLYPURGE requests
     if (req.request == "FASTLYPURGE") {
-      error 403 "Forbiddden";
+      set req.http.Fastly-Purge-Requires-Auth = "1";
     }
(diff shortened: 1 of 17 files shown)


spec/test-outputs/apt-integration.out.vcl
```diff
@@ -25,9 +25,9 @@ backend F_apt {
 sub vcl_recv {
 #FASTLY recv
 
-    # Prevent third parties from flushing CDN caches with FASTLYPURGE
+    # Require authentication for FASTLYPURGE requests unless from IP in ACL
     if (req.request == "FASTLYPURGE") {
```
sihugh: the `apt` configs don't have the ip whitelist check.  Is this intentional?  (If so, the comment might need tweaking)

kevindew: Whoops! classic copy + pasting error. Sorted now.

apt - is our own internal aptly server that we pull our software from for our ubuntu boxes.

Thanks for looking @sihugh :+1:

PR #153, opened 2019-05-23
Cache non-GET/HEAD requests with 404/405 responses

bilbof:
Paired with @koetsier on this.

I would appreciate thoughts on this (implementation details, concerns/risks, in general and in response to the P1 issue).

## What does this change do?

At the moment we only cache GET/HEAD requests.

Now we'll cache any request, if that request returns a 404 (Not Found) or 405 (Method Not Allowed).

Note that administration/publishing applications don't sit behind this cache, so (e.g.) requests to create documents won't be affected by this change.

### Example

If a DELETE request is made to a page on the gov.uk website - for the most part not permitted - this should return a 405.

We then cache this response, so future DELETE requests to the same resource will not reach the origin until the default Time-to-live (TTL) for the cached object expires (at the moment, the default TTL is 5000 seconds).

## Motivation for the change

This change is in response to a P1 issue in February 2019. More details of the P1: https://trello.com/c/6P4c8RiN/396 (if you can't access this let me know).

In general, it will reduce the workload for servers behind the cache.

## Explain it to me like I didn't read the VCL documentation

Varnish (our cache), which is run as a managed service by Fastly (our CDN), allows us to modify how we cache responses. This is made possible with VCL ([Varnish Configuration Language](https://varnish-cache.org/docs/2.1/reference/vcl.html#varnish-configuration-language) version 2.1).

This change modifies the VCL we run for www.gov.uk on all environments.

The change to `vcl_recv` (run before we've made a request to the backend servers) means that we lookup all requests (regardless of request method) in the cache. Therefore, any cached POST/DELETE/GET/etc. response that is cached, if requested again, will be served by the cache and won't reach servers behind the cache.

The change to `vcl_fetch` (run after we've got a response from the backend servers) means that we only cache non-GET/HEAD requests if they return a 404 or 405.

## Concerns

These are questions that have come up, which others more experienced with Fastly/Varnish may have thoughts about.

**Is it possible that sensitive data could be cached?**

I think that sensitive data won't be cached. The new cached objects are only 404 / 405 responses, though I might be wrong. We will store hashed requests (including query params) in the same way as GET requests, though I'm not sure if we consider these sensitive.

**Is it possible that this cache logic could prevent valid requests?**

I don't think so, since query params and so on are cached and create unique cache objects. For example: `gov.uk/search?q=breakfast` and `gov.uk/search?q=lunch` will create two unique cache objects. In this case, you won't receive breakfast when you've asked for lunch.

More dangerously, if a DELETE request is made on a resource, which returns a 404, and then the resource is created and the cache is not invalidated, then a subsequent DELETE request might return an invalid 404.

```
DELETE /horse/1
> 404 Not Found (horse doesn't exist)
CREATE /horse { id: 1 }
DELETE /horse/1
> (from cache): 404 Not Found (horse doesn't exist) <- incorrect.
```

**Will backends that incorrectly respond with 404/405 then lead to a confusing debugging experience for developers?**

**Will a higher number of `lookup` calls on the Varnish cache have an adverse affect?**

I suspect not, since we pretty much only see GET requests. The ratio of GET requests to other requests is expected to stay the same.

### DIFF
--- a/vcl_templates/www.vcl.erb
+++ b/vcl_templates/www.vcl.erb
@@ -256,10 +256,6 @@ sub vcl_recv {
 
 #FASTLY recv
 
-  if (req.request != "HEAD" && req.request != "GET" && req.request != "FASTLYPURGE") {
-    return(pass);
-  }
-
   <% template_path = File.join(File.dirname(__FILE__), "vcl_templates/_multivariate_tests.vcl.erb") -%>
   <%= ERB.new(File.new(template_path).read, nil, "-", "_erbout2").result(binding) %>
 
@@ -326,6 +322,21 @@ sub vcl_fetch {
     if (beresp.http.Fastly-Backend-Name ~ "mirrorS3") {
       set beresp.ttl = 900s;
       set beresp.http.Cache-Control = "max-age=900";
+    }
+  }
+
+  # Cache non-GET/HEAD/FASTLYPURGE requests if they return a 404 or 405
+  if (!(req.request == "GET" || req.request == "HEAD" || req.request == "FASTLYPURGE")) {
+    if (http_status_matches(beresp.status, "404,405")) {
+      # Cache these 404/405 responses
+      set beresp.cacheable = true;
+      set beresp.ttl = 10s;
+    } else {
+      # Don't cache these responses.
+      # Among other things, this creates a hit-for-pass cache object for this request
+      # Explanation of what calling pass in vcl_fetch does here:
+      # https://docs.fastly.com/guides/vcl-tutorials/understanding-the-different-pass-action-behaviors#using-a-cache-setting
+      return(pass);
     }
   }
 
--- a/spec/test-outputs/www-integration.out.vcl
+++ b/spec/test-outputs/www-integration.out.vcl
@@ -212,10 +212,6 @@ sub vcl_recv {
 
 #FASTLY recv
 
-  if (req.request != "HEAD" && req.request != "GET" && req.request != "FASTLYPURGE") {
-    return(pass);
-  }
-
     # Begin dynamic section
 if (table.lookup(active_ab_tests, "Example") == "true") {
   if (req.http.User-Agent ~ "^GOV\.UK Crawler Worker") {
@@ -379,6 +375,21 @@ sub vcl_fetch {
     if (beresp.http.Fastly-Backend-Name ~ "mirrorS3") {
       set beresp.ttl = 900s;
       set beresp.http.Cache-Control = "max-age=900";
+    }
+  }
+
+  # Cache non-GET/HEAD/FASTLYPURGE requests if they return a 404 or 405
+  if (!(req.request == "GET" || req.request == "HEAD" || req.request == "FASTLYPURGE")) {
+    if (http_status_matches(beresp.status, "404,405")) {
+      # Cache these 404/405 responses
+      set beresp.cacheable = true;
+      set beresp.ttl = 10s;
+    } else {
+      # Don't cache these responses.
+      # Among other things, this creates a hit-for-pass cache object for this request
+      # Explanation of what calling pass in vcl_fetch does here:
+      # https://docs.fastly.com/guides/vcl-tutorials/understanding-the-different-pass-action-behaviors#using-a-cache-setting
+      return(pass);
     }
   }
 
--- a/spec/test-outputs/www-production.out.vcl
+++ b/spec/test-outputs/www-production.out.vcl
@@ -213,10 +213,6 @@ sub vcl_recv {
 
 #FASTLY recv
 
-  if (req.request != "HEAD" && req.request != "GET" && req.request != "FASTLYPURGE") {
-    return(pass);
-  }
-
     # Begin dynamic section
 if (table.lookup(active_ab_tests, "Example") == "true") {
   if (req.http.User-Agent ~ "^GOV\.UK Crawler Worker") {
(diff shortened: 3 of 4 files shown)


vcl_templates/www.vcl.erb
```diff
@@ -256,10 +256,6 @@ sub vcl_recv {
 
 #FASTLY recv
 
-  if (req.request != "HEAD" && req.request != "GET" && req.request != "FASTLYPURGE") {
-    return(pass);
-  }
```
bilbof: This will mean we now lookup every request in the cache, rather than just GET/HEAD.

vcl_templates/www.vcl.erb
```diff
@@ -327,6 +323,12 @@ sub vcl_fetch {
       set beresp.ttl = 900s;
       set beresp.http.Cache-Control = "max-age=900";
     }
+
+    # Only cache non-GET/HEAD requests if they return a 404 or 405
+    if (!(req.request == "GET" || req.request == "HEAD") && !http_status_matches(beresp.status, "404,405")) {
+      set beresp.ttl = 0s;
+      set beresp.cacheable = false;
+    }
```
bilbof: The intent of this change is to avoid cacheing non-GET/HEAD requests if their response status is not 404,405.

One thing I am not clear on is whether the 405 response will be cached. By the Varnish default 405 responses are not cached. So I don't know if we need something like this in there:
```
    if (http_status_matches(beresp.status, "405")) {
      # cache the 405 response.
    }
```

bilbof: > ### beresp.cacheable
> True if the request resulted in a cacheable response. A response is considered cacheable if HTTP status code is 200, 203, 300, 301, 302, 404 or 410 and pass wasn’t called in vcl_recv. If however, both the TTL and the grace time for the response are 0 beresp.cacheable will be 0.
> beresp.cacheable is writable.

So it sounds like adding the above with `beresp.cacheable = true` would be necessary, as setting the default ttl may not be enough.

I feel like I need to read the source code of Varnish (and Fastly's implementation which is _based on_ Varnish 2.1) to know for sure what is necessary.

vcl_templates/www.vcl.erb
kevindew: Looks like you've also included purge here that you'd want to keep.

vcl_templates/www.vcl.erb
```diff
@@ -327,6 +323,12 @@ sub vcl_fetch {
       set beresp.ttl = 900s;
       set beresp.http.Cache-Control = "max-age=900";
     }
+
+    # Only cache non-GET/HEAD requests if they return a 404 or 405
+    if (!(req.request == "GET" || req.request == "HEAD") && !http_status_matches(beresp.status, "404,405")) {
+      set beresp.ttl = 0s;
+      set beresp.cacheable = false;
```
kevindew: If you're using these two to avoid caching I think you can do `return(pass)`: https://varnish-cache.org/docs/2.1/tutorial/vcl.html#actions

bilbof: Thanks, I had to do some reading to understand why, but that makes sense.

Is it correct that the resulting behaviour of calling `return(pass)` in `vcl_fetch` would be to create a hit-for-pass object, and thereby avoid a `lookup` for this request in future (while the ttl lasts)?

vcl_templates/www.vcl.erb
bilbof: I think `req.resuest != "FASTLYPURGE"` here means that `return(pass);` would not be called for purge requests, so removing the whole block won't have an affect. I added in FASTLYPURGE to the check in vcl_fetch, though.

kevindew:
Great PR description 👏 🥇

bilbof:
@sihugh good question, I'm not sure. Do we have docs for [cache-clearing-service](https://github.com/alphagov/cache-clearing-service)?

vcl_templates/www.vcl.erb
kevindew: Ah sorry my bad

kevindew:
Cache clearing service would clear them, but I'd be super keen that we can try avoid needing to use that.

If possible I'd suggest we don't need to set a ttl directly as hopefully that can be set upstream - I believe we already set 10s on 404 and not sure about 405. If we do need to put a ttl in here I'd suggest not even going beyond 60s as we don't have any use case other than DDOS attack and we can easily handle one of those every 10s.

One thing I wondered is whether it is in scope to protect against different param combinations. 

eg if you did this: 

```
for i in {1..10}; do curl -i --silent -X POST -d param=$RANDOM https://www.gov.uk/404 | grep X-Cache; done
```

I know we don't do the same for GET so we get all misses for: 

```
for i in {1..10}; do curl -i --silent https://www.gov.uk/\?$RANDOM | grep X-Cache; done
```

so could understand it being out of scope.

sihugh:
I thought the latter might be out of scope for now @kevindew, but I agree with capping the ttl.

elliotcm:
> Would this cache be cleared by the cache clearing service? I wonder whether 5000s/default_ttl is too long and it might prove to be annoying at some point.

From what I'm reading here this is requests we rarely see under normal circumstances, so while I agree that caching can lead to confusion, I'd presume that it would minimally affect us during normal day to day, and it's not like we're manually sending these during testing.

bilbof:
@kevindew @elliotcm @sihugh thanks for your comments.

I've put this change up on integration for the rest of the afternoon if you would like to test/tweak it.

I can reduce the TTL below 5000s for these requests. Would 10s be OK?

kevindew:
> @kevindew @elliotcm @sihugh thanks for your comments.
> 
> I've put this change up on integration for the rest of the afternoon if you would like to test/tweak it.
> 
> I can reduce the TTL below 5000s for these requests. Would 10s be OK?

Yup 10s sounds good 👍

kevindew:
Also commit message looks a bit strange with the trello card title

bilbof:
Thanks, I fixed the commit title and changed the TTL to 10s.

PR #156, opened 2019-05-24
Optimize mirror vcl

fredericfran-gds:
# Context

There is no need for:
1. blacklisting IPs because only few IPs are allowed to connect to this
fastly service by default
2. AB tests because no such tests are needed on the mirror fastly service

# Decisions
1. remove blacklist ips section
2. remove import of AB tests


fredericfran-gds:
I use line 147 in vcl_templates/mirror.vcl.erb to whitelist IPs that can access the mirror fastly service

PR #157, opened 2019-06-04
[DO NOT MERGE, testing in staging] new www fastly service vcl to use new mirror buckets

fredericfran-gds:
# Context

New gov.uk mirror redesign:
If the www origin backend is down, fastly will use stale cache as before.
If object is not in cache, the S3 mirror bucket in London is used. If this fails, the
replica S3 mirror bucket in Ireland is used. As last resort,
we use the Google Cloud Storage bucket in multi-region EU.

Other changes:
1. we no longer have a sick backend because it is pointless as we keep the grace period to 24 hours regardless, so the example previously followed: https://book.varnish-software.com/3.0/Saving_a_request.html#example-evil-backend-hack no longer applies. Confirmed with fastly that this is fine.
2. if the backend is origin and we encounter an error from the origin, we try to serve stale too.

# Decisions
1. re-write the www vcl config to accommodate the 3 mirror buckets and removal of mirror machines in Carrenza 

Data component of this PR is in: https://github.com/alphagov/govuk-cdn-config-secrets/pull/102

### DIFF
--- a/deploy-service.sh
+++ b/deploy-service.sh
@@ -2,7 +2,7 @@
 set -eu
 
 rm -rf govuk-cdn-config-secrets
-git clone git@github.com:alphagov/govuk-cdn-config-secrets.git -b add_www_staging_values
+git clone git@github.com:alphagov/govuk-cdn-config-secrets.git
 
 cp govuk-cdn-config-secrets/fastly/fastly.yaml .
 
--- a/deploy-service.sh
+++ b/deploy-service.sh
@@ -2,7 +2,7 @@
 set -eu
 
 rm -rf govuk-cdn-config-secrets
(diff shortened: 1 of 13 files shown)


deploy-service.sh
```diff
@@ -2,7 +2,7 @@
 set -eu
 
 rm -rf govuk-cdn-config-secrets
-git clone git@github.com:alphagov/govuk-cdn-config-secrets.git
+git clone git@github.com:alphagov/govuk-cdn-config-secrets.git -b add_www_staging_values
```
tijmenb: This needs to point to master before merging.

fredericfran-gds: yes, thanks, we know. It is in the trello card before we merge etc

fredericfran-gds:
no longer needed

PR #161, opened 2019-06-14
new production and staging mirror fastly service

fredericfran-gds:
# Context

The `mirror` fastly service is used to test the mirrors for `www` fastly service.
Hence, the `mirror` fastly service will have the same vcl as the `www` fastly service except for:
1. clients should be in the whitelist to access the `mirror` fastly service.

We test the mirrors after deploying the `mirror` fastly service by going into the fastly management console and change the active vcl for the service so that it points to an unreachable origin and therefore the service will fallback to mirrors.

# Decisions
1. copy the future www template vcl to be tested to mirror template vcl and modify the template vcl to restrict client for all environments rather than staging and integration only.

### DIFF
--- a/spec/test-outputs/mirror-integration.out.vcl
+++ b/spec/test-outputs/mirror-integration.out.vcl
@@ -1,9 +1,8 @@
-# Mirror backend for S3
-backend F_mirrorS3 {
-    .connect_timeout = 1s;
+backend F_origin {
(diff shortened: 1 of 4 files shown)


fredericfran-gds:
we just want minimum changes from the www vcl template because we want the mirror service behave like the www service with the exception that we will break the origin to test fallback to mirrors

PR #163, opened 2019-06-18
Add support for TLS 1.3 with 0-RTT

rubenarakelyan:
This commit adds VCL code to support TLS 1.3 with 0-RTT session resumption and prevent 0-RTT for anything other than GET or HEAD requests without query string data, to stop replay attacks.

### DIFF
--- a/spec/test-outputs/www-integration.out.vcl
+++ b/spec/test-outputs/www-integration.out.vcl
@@ -78,6 +78,18 @@ sub vcl_recv {
      error 801 "Force SSL";
   }
 
+  # START: Test code for TLS 1.3 with 0-RTT session resumption
+  # 0rtt (early data accepted) ensure GET|HEAD
+  if (req.http.Early-Data && req.request != "GET" && req.request != "HEAD") {
+    error 725 "TLS Too Early";
+  }
+
+  # 0rtt (early data accepted) ensure GET|HEAD without query params
+  if (req.http.Early-Data && req.url.qs != "") {
+    error 725 "TLS Too Early";
+  }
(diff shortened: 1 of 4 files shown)


Nooshu:
Additional context: Speaking to Fastly about enabling TLS v1.3 as it can improve performance. 0-RTT improves it further. [Cloudflare blog](https://blog.cloudflare.com/introducing-0-rtt/) and [KeyCDN blog](https://www.keycdn.com/blog/tls-1-3-support).

With [TLS v1.3](https://tools.ietf.org/html/draft-ietf-tls-tls13-28), new sessions with have a 1-RTT for negotiation. Resumed sessions will require 0-RTT. 

TLS v1.3 has around [72% browser support](https://caniuse.com/#feat=tls1-3) in the UK at the time of writing.

Nooshu:
1. For many users it won't mean anything at all. What it will stop is the possibility of a replay attack where by an attacker can copy the encrypted 0-RTT data and send it to the server again and pretend to be the same user. Under these conditions a [425 status is returned](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/425) (currently only supported by Firefox).
2. Code has come directly from Fastly and is recommended if enabling 0-RTT.
3. The test is the code itself. Testing the browser to see if it supports the `Early-Data` [header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Early-Data).
4. We can test in FF58+, as it is the only browser at the moment that supports 0-RTT (or at least has it enabled by default)

rubenarakelyan:
Closing for now while we test with just TLS 1.3.

PR #164, opened 2019-06-18
if object is 404, serve govuk 404 body rather than backend 404 body

fredericfran-gds:
# Context

There is a fallback to mirrors if page not found on origin website and if the page is not found in the mirrors, the mirror response body is sent to the client. We actually want the response body to be the govuk customize message.

# Decisions
If an object is 404 and comes from the mirrors, send the govuk 404 custom response body.

### DIFF
--- a/spec/test-outputs/mirror-integration.out.vcl
+++ b/spec/test-outputs/mirror-integration.out.vcl
@@ -338,7 +338,7 @@ sub vcl_error {
     return (deliver);
   }
 
-  if (obj.status == 804) {
+  if ((obj.status == 804) || (obj.status == 404 && req.restarts > 0))
     set obj.status = 404;
     set obj.response = "Not Found";
     set obj.http.Fastly-Backend-Name = "force_not_found";
--- a/spec/test-outputs/mirror-production.out.vcl
+++ b/spec/test-outputs/mirror-production.out.vcl
@@ -500,7 +500,7 @@ sub vcl_error {
     return (deliver);
   }
 
-  if (obj.status == 804) {
+  if ((obj.status == 804) || (obj.status == 404 && req.restarts > 0))
     set obj.status = 404;
     set obj.response = "Not Found";
     set obj.http.Fastly-Backend-Name = "force_not_found";
--- a/spec/test-outputs/mirror-staging.out.vcl
+++ b/spec/test-outputs/mirror-staging.out.vcl
@@ -500,7 +500,7 @@ sub vcl_error {
     return (deliver);
   }
 
-  if (obj.status == 804) {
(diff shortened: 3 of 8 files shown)


benthorner:
Sorry to drag this one back out of the past @fredericfran-gds. Do you remember any more detail about what happened here? My reading of [this comment](https://github.com/alphagov/govuk-cdn-config/blob/d77313abdb5098e2b350de7a0992375e50ff03a3/vcl_templates/www.vcl.erb#L354) is that this shouldn't be necessary.

fredericfran-gds:
I think what this means is we want mirrors to return either 200 or 503 but some mirrors do return other http codes and if they do, we force them to 503 and display a nice message to the public user that we are experiencing technical difficulties.

benthorner:
So this is the interesting bit: do we know when the mirrors might return another error code?

fredericfran-gds:
They send 404 not found because not all stuff are mirrored. For e.g. we put this in place because when licensify was experiencing some issues, users were getting some custom s3 messages back.

benthorner:
Do you have an example of the problem? I'm still not sure why sometimes we return a `503`, but other times an actual `404`.

fredericfran-gds:
I guess mirrors can behave in many ways: e.g. 503 for some internal error, 404 if the data is not there, 401 if authentication was misconfigured. We want all these errors to be 503 because there is a specially handling later to make them be a custom 503 govuk error.

benthorner:
Oh, I think I see now. The comment is talking about "errors from the mirrors" in the sense that "the mirror is broken". So if we failed over to mirrors, and the page wasn't there, we would return a 404 response (from S3), and we'd need this change to render it nicely.

I guess we only saw it for licensify because that has a lot of 5xx issues, so we rely on mirrors more than we do for most of the rest of GOV.UK. Does that sound right?

fredericfran-gds:
Yes, this is correct. Apologies if we did not convey this clearly. Feel free to raise a PR so that the comments can be more clear.

benthorner:
Ah, but now I'm confused again. We never do an "error 404" in our VCL, like [we do for 503](https://github.com/alphagov/govuk-cdn-config/blob/d77313abdb5098e2b350de7a0992375e50ff03a3/vcl_templates/www.vcl.erb#L372), so how do we get into `vcl_error` in the first place?

I tried to reproduce some kind of 404 with licensing apps with the origin down, but I couldn't get it to work. Maybe that's because "backend down" is different to "a single 5xx".

Let me know if there's someone else you'd prefer me to ask 😄

fredericfran-gds:
I updated the description here, it should have referred to a 503 govuk body.

PR #170, opened 2019-07-19
Search cluster AB test

bilbof:
This adds an AB test for search clusters. 95% will go to the A/Default cluster. 5% will go to the new elasticsearch 6 cluster. The expiry for the AB test will be 1 day.

A subsequent PR will activate this test by changing `SearchClusterABTest` to `true` in `configs/dictionaries/active_ab_tests.yaml`.

Follows https://docs.publishing.service.gov.uk/manual/run-ab-test.html#2-how-to-set-up-an-ab-test
Trello: https://trello.com/c/OLcUGWeo/858


karlbaker02:
Unless something has changed recently, I'm not sure why the VCL for the mirrors has also been updated? In previous A/B tests only for the VCL affecting `www-ENV` needed to be updated.

Also, I'd suggest enabling this test with initial percentages for traffic to variants as

```
A: 100%
B: 0%
```

and a short cookie expiry (though 24 hours as you have it is probably fine).

When this PR is merged we'll invalidate the cache, so if we switch it on with `A: 100%` to begin with, we'll reduce load on origin at the point when you send traffic to the `B` variant.

bilbof:
@karlbaker02 

Thanks, I changed the percentages to 100 and 0. Am I correct in thinking that the AB test wont be active until this is merged? https://github.com/alphagov/govuk-cdn-config/pull/171/files#diff-40b85c5bdce3d06be10655c29f214e18R9

The code impacting the mirror VCL was autogenerated. It looks like the AB tests are already present there https://github.com/alphagov/govuk-cdn-config/commit/b2443db2450bfe021f195849034f8416ee92f1ea

karlbaker02:
Yep, that's right Bill. Suggest you probably make it active as part of this PR, then do another PR when you actually want to turn on the A/B test to update the variants.

bilbof:
Thanks, I've changed the test to be active.

PR #180, opened 2019-07-23
Update Fastly IPs

thomasleese:
It looks like Fastly have updated their IP addresses: https://deploy.publishing.service.gov.uk/job/Check_CDN_IP_Ranges/1444/console


benthorner:
Out of interest, is there a definitive list published by Fastly? Aha...

https://api.fastly.com/public-ip-list

PR #199, opened 2019-09-19
Add Landing Page AB test placeholder

vanitabarrett:
PR to add the landing page test. This adds the test, switched off for now, and with a short cookie expiry. When we want to turn the test on, we will need to switch it to on and extend the cookie expiry to the length of the AB test.

Followed these docs: https://docs.publishing.service.gov.uk/manual/run-ab-test.html


ab_tests/ab_tests.yaml
```diff
@@ -13,3 +13,6 @@
 - SearchClusterQueryABTest:
   - A
   - B
+- LandingPageTest:
```
1pretz1: Slightly weird that the naming convention here is inconsistent. Half the AB test names have a `ABtest` suffix and half don't. Not your fault though as there is both patterns before LandingPageTest even existed! 🙁

vanitabarrett: Yeah - I went with the option without ABTest as the docs say the name should be prefixed with `ABTest-` but that this doesn't need to be included in the code as it's already configured in Fastly... the full name being `ABTest-LandingPageABTest` seemed a bit odd to me 🤷‍♀

1pretz1: Yup I agree that the suffix is over kill. If I feel like a good Samaritan / brave enough I may make the naming consistent across all the 4 tests 😆

PR #200, opened 2019-09-24
Make landing page AB test active

vanitabarrett:
This PR sets the landing page test to active, which I think is needed to deploy and be able to check it has been set up properly.

Traffic won't actually start being sent to any variants until we've implemented the logic in Collections and have adjusted the percentages to direct a proportion of traffic to B.

### DIFF
--- a/configs/dictionaries/active_ab_tests.yaml
+++ b/configs/dictionaries/active_ab_tests.yaml
@@ -7,4 +7,4 @@ Example: true
 ViewDrivingLicence: false
 FinderAnswerABTest: true
 SearchClusterQueryABTest: true
-LandingPageTest: false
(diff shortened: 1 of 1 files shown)


configs/dictionaries/active_ab_tests.yaml
```diff
@@ -7,4 +7,4 @@ Example: true
 ViewDrivingLicence: false
 FinderAnswerABTest: true
 SearchClusterQueryABTest: true
-LandingPageTest: false
+LandingPageTest: true
```
kevindew: Should we name this as BrexitLandingPageTest as I'm not sure LandingPage makes this sufficiently distinct

PR #213, opened 2019-10-30
Remove querystrings and sub paths from register-to-vote

bevanloon:
Any paths under our three `register-to-vote` base paths are not
neccessary and we can safely normalise to the base path.

Similarly, we don't need the querystrings on these paths and allowing
them just leads to more needless requests to origin.

[Trello](https://trello.com/c/cpi5iw7T/846-temporarily-strip-cache-bust-string-on-register-to-vote-in-our-vcl)

### DIFF
--- a/spec/test-outputs/www-production.out.vcl
+++ b/spec/test-outputs/www-production.out.vcl
@@ -203,6 +203,10 @@ sub vcl_recv {
     set req.http.original-url = req.url;
   }
 
+  if (req.url.path ~ "(?i)^/register-to-vote(-armed-forces|-crown-servants-british-council-employees)?") {
+    set req.url = std.tolower(re.group.0);
+  }
+
   # Common config when failover to mirror buckets
   if (req.restarts > 0) {
     set req.url = req.http.original-url;
--- a/vcl_templates/www.vcl.erb
+++ b/vcl_templates/www.vcl.erb
@@ -248,6 +248,10 @@ sub vcl_recv {
     set req.http.original-url = req.url;
   }
 
+  if (req.url.path ~ "(?i)^/register-to-vote(-armed-forces|-crown-servants-british-council-employees)?") {
+    set req.url = std.tolower(re.group.0);
+  }
+
   # Common config when failover to mirror buckets
   if (req.restarts > 0) {
     set req.url = req.http.original-url;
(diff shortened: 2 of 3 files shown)


spec/test-outputs/www-production.out.vcl
```diff
@@ -203,6 +203,17 @@ sub vcl_recv {
     set req.http.original-url = req.url;
   }
 
+  if (req.url.path ~ "(?i)^/register-to-vote-armed-forces") {
+    set req.url.path = "/register-to-vote-armed-forces";
```
thomasleese: Do we need to set `req.url.path` if we're setting `req.url` as a whole in the next line?

timblair: I commented on this in the `vcl_templates/www.vcl.erb` change. No, we don't need to set both, and in fact `req.url.path` is read-only, so attempting to set it will fail.

vcl_templates/www.vcl.erb
```diff
@@ -248,6 +248,17 @@ sub vcl_recv {
     set req.http.original-url = req.url;
   }
 
+  if (req.url.path ~ "(?i)^/register-to-vote-armed-forces") {
```
timblair: As all paths have a common format, we can compress this `if` block down to a single clause, and use the regex captures to dictate the canonical path to set:

```
if (req.url.path ~ "(?i)^/register-to-vote(-armed-forces|-crown-servants-british-council-employees)?") {
  set req.url = std.tolower(re.group.0);
}
```

[Here's a demo](https://fiddle.fastlydemo.net/fiddle/48443bdc) using the most-excellent Fiddle tool to show it works.

vcl_templates/www.vcl.erb
```diff
@@ -248,6 +248,17 @@ sub vcl_recv {
...
+    set req.url.path = "/register-to-vote-armed-forces";
```
timblair: `req.url.(basename|dirname|ext|path|qs)` are all read-only, computed values based on `req.path` (see e.g. documentation for [`req.url.path`](https://docs.fastly.com/vcl/variables/req-url-path/). `set req.url = ...` is the only line required, which will both set the canonical path, and remove any querystring.

PR #245, opened 2020-02-03
Remove VCL code for caching non-GET/HEAD requests

bilbof:
This reverts commit 689fe7eb4c800ac9fb4393c150133d1c3285389a.

We found a bug in this code that prevents FASTLYPURGE requests
from working as expected. In the vcl_hash we include the request
method in the hash key we use to store and lookup objects in the cache.
This works fine for GET/HEAD/POST requests because we can cache
the responses to these requests as separate objects (a post response
might be different to a get response).

However, it means that FASTLYPURGE requests lookup items in the cache
using FASTLYPURGE as part of the hash, preventing purges from working
as expected.

Rather than attempting to patch this by adding more complexity to the
VCL, we've decided to approach the problem we were trying to solve in
a different way.

Thanks to @kevindew for working on this with me.

https://trello.com/c/7jFNfgr6/1346

### DIFF
--- a/spec/test-outputs/www-integration.out.vcl
+++ b/spec/test-outputs/www-integration.out.vcl
@@ -121,6 +121,10 @@ sub vcl_recv {
 
 #FASTLY recv
 
+  if (req.request != "HEAD" && req.request != "GET" && req.request != "FASTLYPURGE") {
+    return(pass);
+  }
+
     # Begin dynamic section
 if (table.lookup(active_ab_tests, "Example") == "true") {
   if (req.http.User-Agent ~ "^GOV\.UK Crawler Worker") {
@@ -226,18 +230,6 @@ sub vcl_fetch {
     if (beresp.http.Fastly-Backend-Name ~ "^mirror") {
       set beresp.ttl = 900s;
       set beresp.http.Cache-Control = "max-age=900";
-    }
-  }
-
-  # Cache non-GET/HEAD/FASTLYPURGE requests if they return a 404 or 405
-  if (!(req.request == "GET" || req.request == "HEAD" || req.request == "FASTLYPURGE")) {
-    if (http_status_matches(beresp.status, "404,405")) {
-      # Cache these 404/405 responses
-      set beresp.ttl = 5000s;
-      set beresp.cacheable = true;
-    } else {
-      # Don't cache these responses.
-      set beresp.cacheable = false;
     }
   }
 
@@ -361,21 +353,5 @@ sub vcl_pass {
 }
 
 sub vcl_hash {
-  # include request method
-  set req.hash += req.request;
-
-  # support caching non-GET/HEAD requests
-  if (!(req.request == "GET" || req.request == "HEAD" || req.request == "FASTLYPURGE") && req.postbody) {
-    set req.hash += req.postbody;
-  }
-
-  # if handling GET/HEAD requests, fall back to normal
-  {
-    set req.hash += req.url;
-    set req.hash += req.http.host;
-    set req.hash += "#####GENERATION#####";
-    return (hash);
-  }
-
 #FASTLY hash
 }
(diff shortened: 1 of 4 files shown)


barrucadu:
Will deploying this effectively invalidate the entire cache in one go?

bilbof:
@barrucadu yes it would do. It must've been the same situation when we made this change, so perhaps deploying it early in the morning would work OK?

bilbof:
Ah that's right, thanks Kevin. @barrucadu this code has already been manually added to the production VCL.

> it continued to pull the existing items out of the cache until they expired.

I would've expected changing how the cache key is generated to mean that the objects already in the cache wouldn't be returned to users. Forgive me if I have misunderstood the role of `vcl_hash`!

kevindew:
> Ah that's right, thanks Kevin. @barrucadu this code has already been manually added to the production VCL.
> 
> > it continued to pull the existing items out of the cache until they expired.
> 
> I would've expected changing how the cache key is generated to mean that the objects already in the cache wouldn't be returned to users. Forgive me if I have misunderstood the role of `vcl_hash`!

Me too - It's a risk I hadn't considered when I tried this out on staging, however the experience was one where everything continued cached until the 30 mins expired. I wonder if we can find the source code to see what it uses.

bilbof:
That's an odd one! This `vcl_hash`  is already running in production, so it shouldn't invalidate the cached objects, but I'll check that on staging before it goes out.

 up cached items by their old key would work
According to [this doc](https://varnish-cache.org/docs/trunk/users-guide/vcl-hashing.html) (might not be our version) this is the default vcl:

```
sub vcl_hash {
    hash_data(req.url);
    if (req.http.host) {
        hash_data(req.http.host);
    } else {
        hash_data(server.ip);
    }
    return (lookup);
}
```

So the default hash for an object is `default_hash = url + (host || ip)`, but we changed it to `default_hash + request_method + some other stuff` (I'm not clear what the role of the stuff within the `{ }` block within `vcl_hash` is, or when it gets called).

Also, Fastly add the line `set req.hash += req.vcl.generation;` to the `vcl_hash` code:

```
#--FASTLY HASH BEGIN
# support purge all
set req.hash += req.vcl.generation;
#--FASTLY HASH END
```

kevindew:
> That's an odd one! This `vcl_hash` is already running in production, so it shouldn't invalidate the cached objects, but I'll check that on staging before it goes out.

Same config as prod is on staging too - but we can always roll this back there so you can do some testing.

> I had a look at the VCL, but I couldn't figure out how looking up cached items by their old key would work.
> 
> According to [this doc](https://varnish-cache.org/docs/trunk/users-guide/vcl-hashing.html) (might not be our version) this is the default vcl:
> 
> ```
> sub vcl_hash {
>     hash_data(req.url);
>     if (req.http.host) {
>         hash_data(req.http.host);
>     } else {
>         hash_data(server.ip);
>     }
>     return (lookup);
> }
> ```
> 
> So the default hash for an object is `default_hash = url + (host || ip)`, but we changed it to `default_hash + request_method + some other stuff` (I'm not clear what the role of the stuff within the `{ }` block within `vcl_hash` is, or when it gets called).
> 
> Also, Fastly add the line `set req.hash += req.vcl.generation;` to the `vcl_hash` code:
> 
> ```
> #--FASTLY HASH BEGIN
> # support purge all
> set req.hash += req.vcl.generation;
> #--FASTLY HASH END
> ```

I've done a bit of digging and it transpires that `hash_data` is a function that now replaces the `set req.hash +=` call (https://serverfault.com/a/548588) though code looks almost the same in 2.1: https://varnish-cache.org/docs/2.1/reference/vcl.html#examples

I'm guessing that req.hash is a string so I am struggling to imagine why these items continued to be looked up from the cache after the change as it should have. As I'm assuming that when we do `return(lookup)` that makes a call to vcl_hash.

Probably worth trying this out on staging just to satisfy (or not) our sanities.

bilbof:
For future readers, Kevin and I had a look into this and confirmed that changing how the cache key is generated will mean that the objects already in the cache wouldn't be returned to users.

PR #250, opened 2020-02-27
Add Electric Car AB test

sihugh:
Our variants have equal weight.
We'll extend the cookie age from 1 day when we're confident that things are working OK.

https://trello.com/c/In6XcrQU/1399-set-up-dft-multi-variant-a-b-test-test-dev-work


configs/dictionaries/electric_car_percentages_ab_test.yaml
```diff
@@ -0,0 +1,9 @@
+---
+A: 1
```
bilbof: I think this file needs to be called `electriccarabtest_percentages.yaml` since the dictionary in the generated VCL code has `table.lookup(electriccarabtest_percentages, "A")` - I ran into the same issue earlier in the week :) 

Also, am I right in thinking 1% of traffic is intended for each variant?

bilbof: Also, do you need to add in `Z` here as a variant, since it's expected as the control variant in the AB test in finder-frontend?

sihugh: I've added Z for clarity (I think it's a good idea!), though I don't think it's strictly necessary as we won't ever assign anyone to it here - it's just to pick up the traffic not in the test.

I've added some tests to ensure that we make the correct files and at least reference each test in every file.

PR #251, opened 2020-02-27
[spike] Adopt varnishtest for testing VCL

bilbof:
**Note:** This PR is not ready for merge.

This spike demonstrates how we might use [varnishtest](https://varnish-cache.org/docs/2.1/reference/varnishtest.html) to unit test the VCL for GOV.UK.

I propose that we would generate another instance of VCL for each environment for testing, e.g. www-integration.test.vcl, which we could then run unit tests against.

The tests can be run as part of our CI test suite, with a single command:

```
docker run -v $(pwd):/work rhargreaves/varnishtest /test-vcl.sh spec/unit/www_tests.vtc spec/test-outputs/www-integration.out.vcl
```

I would recommend reading the PR by commit.

Below I discuss the pros and cons of this approach, I'd be grateful for feedback on this work-in-progress.

Pros:

- We will gain more confidence in the VCL we write, enabling us to work in a TDD way.
- This repository will conform to the GDS way, since it will have tests as part of the CI.
- We will catch some bugs before they go to any environment including syntax errors and functional errors.
- Varnishtest is still supported and I found it to be [very well documented](https://varnish-cache.org/docs/trunk/reference/vtc.html) with [lots of example tests](https://github.com/varnishcache/varnish-cache/tree/2.1/bin/varnishtest/tests).
- We will be able to write vanilla VCL, not dependent on Fastly, making it easier to migrate to another CDN.
- This approach could also be applied to the Varnish instances we run on cache machines.
- A side effect of not using Fastly VCL features is that we can deploy CDN changes in one step, not two, as we won't need to use dictionaries. Another side effect is that we will be able to deprecrate the cdn-secrets repo and move those secrets to the main secrets repository.
- Finally, as a result of the above, we will reduce the number of incidents related to changes we make to our VCL config in Fastly, provided the test coverage is good.

Cons:

- This approach requires us to not use Fastly VCL features, or at least not test those we use. See the previous commit for detail. This is because Fastly VCL is not valid VCL so cannot be compiled outside of Fastly's environment.
- Since Fastly runs a customized version of Varnish that this tests against, these tests may build false confidence in the effect of a change.
- The tests are written in VTC, a nice language, but not something folks working on GOV.UK will likely have experience with.
- It requires a very careful initial migration away from using Fastly's VCL - one that itself could cause an incident
- It might encourage more complex logic being implemented in VCL.

Other approaches evaluated

Relying more heavily on Smokey for testing VCL config is an option. The tests we write for VCL behaviour against a live environment would likely become quite flaky.

I looked at using Fastly Fiddle in CI as proposed by Fastly in https://www.fastly.com/blog/testing-fastly-ci. I decided against this route partly because it's an experimental tool not recommended for production

I also considered reviving a previous project for testing VCL that former colleagues worked on:
https://technology.blog.gov.uk/2014/10/01/cdn-acceptance-testing/.

Another recommendation was https://github.com/instrumenta/conftest.

I decided to go for varnishtest since it seemed like the best way to comprehensively test VCL behaviour.

https://trello.com/c/gS8EksUN/1373-investigate-fastly-vcl-unit-tests

The next step for this will be to gather some feedback on the approach and decide whether we want to try out other options or press on with this.

### DIFF
--- a/spec/test-outputs/www-integration.out.vcl
+++ b/spec/test-outputs/www-integration.out.vcl
@@ -75,9 +75,7 @@ sub vcl_recv {
   
 
   # Force SSL.
-  if (!req.http.Fastly-SSL) {
-     error 801 "Force SSL";
-  }
+  # TODO: Wrap this in a for_fastly? if statement
 
   # Serve a 404 Not Found response if request URL matches "/autodiscover/autodiscover.xml"
   if (req.url ~ "(?i)/autodiscover/autodiscover.xml$") {
@@ -91,6 +89,10 @@ sub vcl_recv {
   # chosen
   set req.backend = F_origin;
   set req.http.Fastly-Backend-Name = "origin";
+
+  # TODO: wrap in if statement for test mode %>
+  # uses the default test backend
+  set req.backend = s1;
 
   # Set header to show recommended related links for Whitehall content. This is to be used
   # as a rollback mechanism should we ever need to stop showing these links.
--- a/spec/test-outputs/www-integration.out.vcl
+++ b/spec/test-outputs/www-integration.out.vcl
@@ -1,23 +1,14 @@
 backend F_origin {
     .connect_timeout = 5s;
-    .dynamic = true;
     .port = "443";
-    .host = "foo";
+    .host = "127.0.0.1";
     .first_byte_timeout = 15s;
     .max_connections = 200;
     .between_bytes_timeout = 10s;
-    .share_key = "123";
-
-    .ssl = true;
-    .ssl_check_cert = always;
-    .min_tls_version = "1.2";
-    .ssl_cert_hostname = "foo";
-    .ssl_sni_hostname = "foo";
-
     .probe = {
         .request =
             "HEAD / HTTP/1.1"
-            "Host: foo"
+            "Host: 127.0.0.1"
             "User-Agent: Fastly healthcheck (git version: )"
             "Connection: close";
         .threshold = 1;
@@ -59,23 +50,25 @@ acl purge_ip_whitelist {
   "167.82.224.0"/20;  # Fastly cache node
 }
 
-
-acl allowed_ip_addresses {
-  
+# TODO: Inject ip_address_blacklist at build time dependent on environment
+acl ip_address_blacklist {
 }
+
+
+# acl allowed_ip_addresses {}
 
 
 sub vcl_recv {
 
   # Require authentication for FASTLYPURGE requests unless from IP in ACL
-  if (req.request == "FASTLYPURGE" && client.ip !~ purge_ip_whitelist) {
+  if (req.request == "FASTLYPURGE" && !(client.ip ~ purge_ip_whitelist)) {
     set req.http.Fastly-Purge-Requires-Auth = "1";
   }
 
   
 
   # Check whether the remote IP address is in the list of blocked IPs
-  if (table.lookup(ip_address_blacklist, client.ip)) {
+  if (client.ip ~ ip_address_blacklist) {
     error 403 "Forbidden";
   }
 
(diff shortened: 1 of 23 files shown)


spec/test-outputs/www-integration.out.vcl
```diff
@@ -1,23 +1,14 @@
...
     .first_byte_timeout = 15s;
     .max_connections = 200;
     .between_bytes_timeout = 10s;
-    .share_key = "123";
-
-    .ssl = true;
-    .ssl_check_cert = always;
-    .min_tls_version = "1.2";
-    .ssl_cert_hostname = "foo";
-    .ssl_sni_hostname = "foo";
```
bilbof: These seem to be Fastly-specific. One solution would be to wrap lines like this in an `<% unless testing %>` block so that the VCL will compile for varnishtest. However, they would need to be manually tested, perhaps with [Fastly Fiddle](https://fiddle.fastlydemo.net/).

spec/test-outputs/www-integration.out.vcl
```diff
@@ -59,35 +50,35 @@ acl purge_ip_whitelist {
...
 
   # Force SSL.
-  if (!req.http.Fastly-SSL) {
-     error 801 "Force SSL";
-  }
+  # TODO: Wrap this in a for_fastly? if statement
 
   # Serve a 404 Not Found response if request URL matches "/autodiscover/autodiscover.xml"
-  if (req.url.path ~ "(?i)/autodiscover/autodiscover.xml$") {
+  if (req.url ~ "(?i)/autodiscover/autodiscover.xml$") {
```
bilbof: I've added a test to demonstrate that this behaviour still works.

spec/test-outputs/www-integration.out.vcl
```diff
@@ -99,6 +90,10 @@ sub vcl_recv {
   set req.backend = F_origin;
   set req.http.Fastly-Backend-Name = "origin";
 
+  # TODO: wrap in if statement for test mode %>
+  # uses the default test backend
+  set req.backend = s1;
```
bilbof: This is a test origin backend used by varnishtest - there would be an `<% if test env %>` block wrapping it here.

spec/test-outputs/www-integration.out.vcl
```diff
@@ -137,26 +133,33 @@ if (req.http.Cookie ~ "cookies_policy") {
...
+        # declare local var.denominator_Example_B INTEGER;
+        # declare local var.nominator_Example_B INTEGER;
+        # set var.nominator_Example_B = std.atoi(table.lookup(example_percentages, "B"));
+        # set var.denominator_Example += var.nominator_Example_B;
+        # set var.denominator_Example_A = var.denominator_Example;
+        # if (randombool(var.nominator_Example_A, var.denominator_Example_A)) {
+        #   set req.http.GOVUK-ABTest-Example = "A";
+        # } else {
+        #   set req.http.GOVUK-ABTest-Example = "B";
+        # }
```
bilbof: I've commented this part out because it relies heavily on Fastly syntax (`declare local var`, `table.lookup`, `randombool`) which we'd need to replace but wasn't necessary in this spike.

vcl_templates/_multivariate_tests.vcl.erb
```diff
@@ -5,11 +5,11 @@
 # Begin dynamic section
 <% if ab_tests -%>
 if (req.http.Cookie ~ "cookies_policy") {
-  if (req.http.Cookie:cookies_policy ~ "%22usage%22:true") {
+  if (req.http.Cookie ~ "cookies_policy=%22usage%22:true") {
```
barrucadu: Unless this means "get the `cookies_policy` cookie value and see if it has `%22usage%22:true` as a substring, it's not correct.

bilbof: Yep I would need to add for that 😉

PR #259, opened 2020-03-24
Match from start of path with ^/search

erino:


vcl_templates/www.vcl.erb
```diff
@@ -230,7 +230,7 @@ sub vcl_recv {
   }
 
   # Serve a 503 Service Unavailable response if request URL matches "/search*"
-  if (req.url.path ~ "(?i)/search.*$") {
+  if (req.url.path ~ "(?i)^/search.*$") {
```
thomasleese:
```suggestion
  if (req.url.path ~ "(?i)^/search") {
```

I don't think we need this bit anymore.

PR #266, opened 2020-05-06
Remove /static/a endpoint

kevindew:
The code for this was removed from static in
https://github.com/alphagov/static/pull/2152 and thus this endpoint is
not used anymore. The 802 status response in this host was only used by
this endpoint so can also be removed.


kevindew:
> Interesting use of status codes - I think it's safe to assume none of our apps are returning these codes!

Ah ha yes. It's a common technique in our varnish files to set custom status codes to do custom responses. It ends up a bit hacky - an 802 in this file means something totally different as a 802 in some other files.

PR #271, opened 2020-07-21
Extend default rake task with rubocop linting

barrucadu:
[Trello card](https://trello.com/c/5Gob6WUL/147-for-everyone-ensure-that-rubocop-runs-as-part-of-bundle-exec-rake-as-one-of-the-default-tasks)

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -5,4 +5,9 @@ require "rspec/core/rake_task"
 RuboCop::RakeTask.new
 RSpec::Core::RakeTask.new
 
-task default: %i[rubocop spec]
+desc "Lint Ruby"
+task :lint do
+  sh "bundle exec rubocop --format clang"
+end
+
+task default: %i[rubocop spec lint]


Rakefile
```diff
@@ -5,4 +5,9 @@ require "rspec/core/rake_task"
 RuboCop::RakeTask.new
 RSpec::Core::RakeTask.new
 
-task default: %i[rubocop spec]
+desc "Lint Ruby"
+task :lint do
+  sh "bundle exec rubocop --format clang"
+end
+
+task default: %i[rubocop spec lint]
```
kevindew, line 13:
```suggestion
task default: %i[rubocop spec rubocop]
```

Looks like we've already got the rubocop task available so can simplify this a bit.

ChrisBAshton, line 13: Doesn't this run rubocop twice?

kevindew: Oops just notice we already do rubocop at the start 😆  - I think we can just close this PR looks like it's already done.

barrucadu: How did I not notice that 🤯

PR #285, opened 2020-09-04
Set a 30 second interval for bouncer healthchecks

cbaines:
There are so many requests, I'm hoping this will reduce the
number. I'm not sure about the usefulness of the Fastly healthcheck
for bouncer, as I'm not sure what it should do differently if the
healthcheck fails...


sarahseewhy (approved):
LGTM

PR #288, opened 2020-09-08
Turn off Transition Checker 4  AB Test

hannako:
Turning off the latest Transition Checker AB test because we now have enough data. 

Trello: https://trello.com/c/dCnm694O/435-a-b-test-4-switch-off-and-results


sihugh (approved):
👍  I guess we're going to flip to 🍏  before turning it off?

PR #296, opened 2020-10-16
Turn off ab test 3

hannako:
Turn off transition urgency AB Test 3 as this test is complete.

Related frontend changes https://github.com/alphagov/collections/pull/1988


sihugh (approved):
👍

PR #302, opened 2020-11-11
Reset proxy headers to avoid spoofing alerts

benthorner:
https://trello.com/c/7qWgBlA7/617-investigate-handling-actiondispatchremoteipipspoofattackerror

Previously a user could manually specify these headers, which could
result in a spoofing error for Rails apps trying to detect the client
IP address [1]. Although the use of the headers by the end user is
valid, we can't trust and shouldn't care about any proxies that the
request has been through before reaching our network. This ensures
the proxy headers are reset before the request continues.

Trying this manually (before/after) shows that:

- Fastly does not have a default value for the Client-IP header, so
it's consistent to enforce that it is unset.

- Fastly pre-populates X-Forwarded-For with the "actual" client IP,
so we should preserve that (but ignore any pre-existing header).

Note that HTTP headers are not case sensitive.

[1]: https://github.com/alphagov/govuk_app_config/pull/171

### DIFF
--- a/spec/test-outputs/www-integration.out.vcl
+++ b/spec/test-outputs/www-integration.out.vcl
@@ -105,8 +105,12 @@ sub vcl_recv {
 
   
 
-  # Unspoofable original client address.
+  # Unspoofable original client address (e.g. for rate limiting).
   set req.http.True-Client-IP = req.http.Fastly-Client-IP;
+
+  # Reset proxy headers at the boundary to our network.
+  unset req.http.Client-IP;
+  set req.http.X-Forwarded-For = req.http.Fastly-Client-IP;
 
   # Set a TLSversion request header for requests going to the Licensify application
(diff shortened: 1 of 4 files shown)


benthorner:
Note that to test this I had to manually tweak the Nginx config (`/etc/nginx/nginx.conf`) on one of the `cache` machines, so that it would print the "Client-IP" header:

```
    log_format timed_combined '$remote_addr - $remote_user [$time_local]  '
      '"$request" $status $body_bytes_sent '
      '"$http_referer" "$http_user_agent" '
      '$request_time $upstream_response_time '
      '$gzip_ratio $sent_http_x_cache $sent_http_location $http_host '
      '$ssl_protocol $ssl_cipher '
      '$http_x_forwarded_for '
      '$http_client_ip';
```

kevindew (approved):
Great stuff, nice one on testing it out for real on the machines.

PR #304, opened 2020-11-17
Set urgency AB test to be 100% on B

sihugh:
This precedes updating the app to reflect this.

https://trello.com/c/WOlLI7cG/599-remove-urgency-a-b-test-on-checker-results-page


hannako (approved):
Guessing the specs don't check percentages?

sihugh:
Yep, the specs just check the VCL, and the VCL just reads the config values from a dictionary, so no changes needed.

PR #307, opened 2020-11-19
Fix ability to synthesise redirects from vcl_recv.

sengi:
7e78be3 didn't work for two reasons: firstly, accidentally omitted the vcl_synth subroutine; secondly, Fastly doesn't support vcl_synth() anyway so we need to use vcl_error().

Use vcl_error() to allow synthesising 302 responses from within vcl_recv. We're using this in alphagov/govuk-cdn-config-secrets#131.

Co-authored-by: @OllieJC

Tested: tested positive and negative cases in staging

See also alphagov/govuk-cdn-config-secrets#132

### DIFF
--- a/spec/test-outputs/www-integration.out.vcl
+++ b/spec/test-outputs/www-integration.out.vcl
@@ -328,6 +328,15 @@ sub vcl_error {
     return (deliver);
   }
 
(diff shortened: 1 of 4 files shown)


fredericfran-gds (approved):
I guess since it has been tested with positive and negative cases, there should be no harm.

PR #308, opened 2020-11-19
Allow private VCL code snippets in all environments.

sengi:
The (per-environment) config item `private_extra_code_in_vcl_recv` allows inserting snippets of VCL from the private govuk-cdn-config-secrets repo into `vcl_recv()`.

Allow this to be used in any environment rather than just staging, now that it's been tested in staging.

Also be a bit more flexible by ignoring the case when `private_extra_code_in_vcl_recv` is not specified in the config, instead of failing.

Follows on from #307 and alphagov/govuk-cdn-config-secrets#132.

Test/rollout plan: rspec is passing. Will test in staging once merged, before rolling to prod.

### DIFF
--- a/spec/test-outputs/www-integration.out.vcl
+++ b/spec/test-outputs/www-integration.out.vcl
@@ -86,6 +86,9 @@ sub vcl_recv {
      error 801 "Force SSL";
   }
 
+  
+  # some private vcl code
+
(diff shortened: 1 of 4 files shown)


OllieJC:
Good shout on setting a default 👍

brucebolt (approved):

PR #310, opened 2020-12-04
Add cookieless test

karlbaker02:
This PR adds a new `CookielessAATest`, which is using [the recommended approach from Fastly](https://www.fastly.com/blog/ab-testing-edge) for running A/B tests without cookies, instead relying on an MD5 hash of a client IP address and user-agent info.

There are three variants to this A/A test - `A`, `B` and `Z`. `A` and `B` are the variants used when a user has not yet accepted cookies and will be computed by Fastly on every request using the approach outlined above and will be used to send hit-only data to analytics. The variant `Z` is for when a user has already accepted cookies, at which point we no longer want to send any hit data to analytics for this test.

Due to the fact that this test does not use cookies, the configuration has broken out of the standard test set-up and has defined some additional logic within the `multivariate_tests` partial.


vcl_templates/_multivariate_tests.vcl.erb
```diff
@@ -4,58 +4,112 @@
...
+      set req.http.GOVUK-ABTest-<%= test %> = "<%= variants.last %>";
+<% variants.each do |variant| -%>
+    } else if (req.url ~ "[\?\&]ABTest-<%= test %>=<%= variant %>(&|$)") {
+      # Some users, such as remote testers, will be given a URL with a query string
+      # to place them into a specific bucket.
+      set req.http.GOVUK-ABTest-<%= test %> = "<%= variant %>";
+<% end -%>
+    } else {
+      declare local var.denominator_<%= test %> INTEGER;
+      set req.http.Client-Bucket-Full = digest.hash_md5(client.ip req.http.User-Agent);
```
bilbof, line 77: > When a request is first handled by a Fastly data center, client.ip represents the IP of the end user. In configurations that use shielding, however, the value of client.ip when the request is processed by the shield data center will be the IP of the edge data center. [source](https://developer.fastly.com/reference/vcl/variables/client-connection/client-ip/)

We don't have shielding enabled, but we might in future, so you may want to use `digest.hash_md5(req.http.Fastly-Client-IP req.http.User-Agent` if this goes beyond a spike.

vcl_templates/_multivariate_tests.vcl.erb
```diff
@@ -4,58 +4,112 @@
...
+      set req.http.Client-Bucket-Trimmed = regsub(req.http.Client-Bucket-Full, "^([a-fA-F0-9]{10}).*$","\1");
+      set req.http.Client-Bucket = std.strtol(req.http.X-ClientIDHash-trimmed, 16);
```
bilbof, line 79: I think needs to be `std.strtol(req.http.Client-Bucket-Trimmed, 16);` given you renamed the variable above.

vcl_templates/_multivariate_tests.vcl.erb
```diff
@@ -4,58 +4,112 @@
...
+      declare local var.denominator_<%= test %>_<%= variant %> INTEGER;
+      declare local var.nominator_<%= test %>_<%= variant %> INTEGER;
+      set var.nominator_<%= test %>_<%= variant %> = std.atoi(table.lookup(<%= test.downcase %>_percentages, "<%= variant %>"));
+      set var.denominator_<%= test %> += var.nominator_<%= test %>_<%= variant %>;
+<% end -%>
+<% variants.each_with_index do |variant, idx| -%>
+<% if variant == variants.first -%>
+      set var.denominator_<%= test %>_<%= variant %> = var.denominator_<%= test %>;
+<% elsif variant != variants.last -%>
+<% previous_variant = variants[idx - 1] -%>
```
bilbof, line 90: Nit. The indentation is a bit confusing here - perhaps you could indent it to the same level as `set var.denominator_` so it's clear it's also in the if block?

vcl_templates/_multivariate_tests.vcl.erb
karlbaker02: Thanks Bill, I wasn't aware of this 👍

vcl_templates/_multivariate_tests.vcl.erb
karlbaker02: 👍

vcl_templates/_multivariate_tests.vcl.erb
karlbaker02: I've tried this, but unfortunately the indents affect the following line (which ends up over-indented as a result).

vcl_templates/_multivariate_tests.vcl.erb
```diff
@@ -4,58 +4,112 @@
...
 <% end -%>
 <% end # variants.each-%>
-      }
     }
+<% end # if CookielessAATest -%>
+  }
 <% end # test_config.each -%>
 <% end # ab_tests.each -%>
+}
+else if (req.http.Cookie !~ "cookies_preferences_set") {
```
bilbof, line 62: Is this `else if` (rather than `else`) to enable users to also reject cookieless AB tests via the cookie banner?

bilbof (approved):
LGTM

vcl_templates/_multivariate_tests.vcl.erb
karlbaker02: The reason it's an `else if` is because we only want to place the user in a variant of this test if they haven't yet consented or declined cookies (via the main consent button in the header _or_ via the cookie preferences page). Once they've done either of these two actions, we no longer want this user to participate in the test.

If this statement were an `else` then we would continue to opt them in to the test regardless of whether they've set preferences, which isn't what we want to do. Appreciate that this logic isn't exactly straightforwards, it has also made me do a few double-takes.

bilbof: Thanks, this makes sense 👍

PR #311, opened 2020-12-10
Set GOV.UK-Request-Id to allow request tracing

erino:
To enable request tracing for GOV.UK we need to set a request id at the CDN so we can trace requests through all layers of GOV.UK


karlbaker02 (approved):
Looks good 👍

PR #312, opened 2020-12-14
Parse and normalise URL query for local restrictions page

bilbof:
When a user visits /find-coronavirus-local-restrictions we will remove all query parameters except for the postcode query parameter, which we will normalise (lowercase, restrict to alphanumerics). See [Fastly Fiddle for demonstration](https://fiddle.fastlydemo.net/fiddle/35784abb).


kevindew (approved):
Nice one @bilbof - thanks for putting this together.

I think we'll plan to replace the postcode string replacement with a JS approach in the coming days so that part might not be in this for too long

PR #316, opened 2021-01-06
Recognise 307 responses as cacheable on www.gov.uk 

kevindew:
Trello: https://trello.com/c/HzeMkOac/1038-return-a-temporary-cachable-redirect-for-the-postcode-checker-while-its-unavailable

GOV.UK responds with 307 for some responses to indicate a temporary
redirect. Fastly doesn't currently support them but suggested this
change.

I decided not to include the similar 303 response in here as this seemed
a bit more contentious given it is expected to originate from a POST
request: varnishcache/varnish-cache#3340 and I
don't believe we have a need.

This same change could be applied to our other VCL files but I'm not
aware of them having a need currently so it seems superfluous to
duplicate this for each.

### DIFF
--- a/vcl_templates/www.vcl.erb
+++ b/vcl_templates/www.vcl.erb
@@ -408,6 +408,12 @@ sub vcl_fetch {
     return (pass);
   }
 
+  # Fastly doesn't recognise 307 as cacheable by default as it is based on an
+  # old version of Varnish that also lacked 307 support.
+  if (beresp.status == 307) {
+    set beresp.cacheable = true;
+  }
+
   if (beresp.http.Expires || beresp.http.Surrogate-Control ~ "max-age" || beresp.http.Cache-Control ~"(s-maxage|max-age)") {
     # keep the ttl here
   } else {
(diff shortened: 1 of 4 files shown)


vcl_templates/www.vcl.erb
```diff
@@ -357,6 +357,12 @@ sub vcl_fetch {
 
   set beresp.http.Fastly-Backend-Name = req.http.Fastly-Backend-Name;
 
+  # Fastly doesn't recognise 307 as cacheable by default as it is based on an
+  # old version of Varnish that also lacked 307 support.
+  if (beresp.status == 307) {
+    set beresp.cacheable = true;
```
bilbof, line 363: Perhaps we should also check `beresp.http.Cache-Control` to check if the response should be cached, just in case. What do you think?

We could either add checks like `&& beresp.http.Cache-Control !~ "private" && ...` to the conditional or move this block down to line 417.

kevindew: Thanks Bill. I don't think we need to check the cache control headers as there are already conditionals for that which will result in a pass.

Moving this code to be below those checks, as you suggest, sounds like a good idea though to resolve some confusion about whether `beresp.cacheable` can impact those pass responses.

bilbof (approved):

vcl_templates/www.vcl.erb
bilbof: Ah thanks for clarifying that :)

PR #320, opened 2021-02-23
Allow UI declared backends and conditions to persist on deploy

edwardkerry:
Fastly allows backends and conditions to be set in the UI,
but subsequent VCL deploys removed these.

https://github.com/alphagov/govuk-cdn-config/blob/1c272cfc79ccb3f9fa3ff89325b70a41df0b9a00/deploy_vcl#L38

We now wish to use some Fastly features that are only
available with UI declared backends and associated conditions.


thomasleese (approved):
It's making me wonder if we could just remove this whole "feature" of the deploy process, but this change is fine too.

PR #321, opened 2021-02-23
Allow UI service backends to persist on deploy

edwardkerry:
Fastly allows backends to be set in the UI, but subsequent VCL deploys removed these.

https://github.com/alphagov/govuk-cdn-config/blob/1c272cfc79ccb3f9fa3ff89325b70a41df0b9a00/deploy_vcl#L38

We now wish to use some Fastly features that are only available with UI declared backends and associated conditions.

This PR also re-adds these to the Bouncer deploy job which were accidentally
removed in https://github.com/alphagov/govuk-cdn-config/commit/ac468c955de96d0808c4f1ce8fa7b3bf804f0380


thomasleese (approved):
That's some interesting duplication :joy:

PR #328, opened 2021-03-17
Allow UI compression settings to persist on deploy

richardTowers:
Fastly allows compressions settings to be toggled in the UI. We'd like
to investigate enabling this on staging, and if it looks good roll it
out to production.

The Default compression policy says:

> Compress text to speed the transfer of data. By default we compress:
css, js, html, eot, ico, otf, ttf, json, and svg formats.

> We support Brotli (the default) and gzip compression formats. The
setting depends upon browser support.

Toggling this and looking at the config diff shows that it puts
everything in the `gzip` section of the config (a bit badly named, since
it supports Brotli as well as GZip, but I imagine that's historical).

Currently our deployment script deletes any `gzip` objects it finds in
the UI, so we can't use the feature. Removing this will allow us to
snowflake the config in.

In the long run we should use terraform for this instead.

### DIFF
--- a/lib/deploy_service.rb
+++ b/lib/deploy_service.rb
@@ -60,7 +60,7 @@ private
   def delete_ui_objects(service_id, version_number)
     # Delete objects created by the UI. We want VCL to be the source of truth.
     # Most of these don't have real objects in the Fastly API gem.
-    to_delete = %w[healthcheck cache_settings request_settings response_object header gzip]
+    to_delete = %w[healthcheck cache_settings request_settings response_object header]
     to_delete.each do |type|
       type_path = "/service/#{service_id}/version/#{version_number}/#{type}"
       @fastly.client.get(type_path).map { |i| i["name"] }.each do |name|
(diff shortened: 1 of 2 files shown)


lib/deploy_service.rb
```diff
@@ -60,7 +60,7 @@ def get_dev_version(configuration)
   def delete_ui_objects(service_id, version_number)
     # Delete objects created by the UI. We want VCL to be the source of truth.
     # Most of these don't have real objects in the Fastly API gem.
-    to_delete = %w[healthcheck cache_settings request_settings response_object header gzip]
+    to_delete = %w[healthcheck cache_settings request_settings response_object header]
```
thomasleese, line 63: Something tells me the length of this array is going to continue to get smaller and smaller 😆

thomasleese (approved):

richardTowers:
Closing this - we've managed to enable brotli without using the setting in the UI, so this is no longer needed.

PR #329, opened 2021-03-30
Add Account Bank Holidays test

huwd:
The GOV.UK Accounts team is running a discovery to explore
personalisation of GOV.UK. We are starting with a light touch, entirely
client side offer for users to remember their choice of bank holiday
nation.

We would like to introduce this gradually and so will stagger the number
of users who get offered this choice as we iterate and further test the
feature


configs/dictionaries/accountbankhols_percentages.yaml
```diff
@@ -0,0 +1,3 @@
+---
+A: 50
```
huwd, line 2: WIP PA to confirm

configs/dictionaries/accountbankhols_percentages.yaml
```diff
@@ -0,0 +1,3 @@
...
+B: 50
```
huwd, line 3: WIP PA to confirm

configs/dictionaries/accountbankhols_percentages.yaml
barrucadu: Maxwell has said that 50/50 is a good split.

barrucadu: And now Daisy has said a 20/80 split, which Maxwell has agreed to.

barrucadu (approved):

PR #330, opened 2021-04-08
Add 302 redirect for security.txt in staging

OllieJC:
This is for the security.txt file redirect to a PaaS app which has the `vdp.cabinetoffice.gov.uk` domain:
https://github.com/alphagov/security.txt

PR #331, opened 2021-04-19
Add 302 redirect for security.txt (all envs)

OllieJC:
Following on from https://github.com/alphagov/govuk-cdn-config/pull/330 which is deployed in staging:
https://www.staging.publishing.service.gov.uk/.well-known/security.txt
This change enables the redirect for the security.txt file in the other environments including prod.


sihugh (approved):
🚀

PR #332, opened 2021-04-23
Add brotli vcl config

MuriloDalRi:
Makes use of increased compression, reducing download size for users when browsing GOV.UK

https://trello.com/c/u9qiWQr1/2476-3-enable-brotli-compression-in-fastly-for-production


edwardkerry (approved):
I think this is in the correct place!

Nooshu (approved):
Yes this looks great. Thanks @MuriloDalRi

PR #336, opened 2021-05-05
Add normalization vcl for Brotli support

Nooshu:
Feedback from Fastly support:
> It looks like `www.gov.uk` doesn't have the normalization vcl in vcl_recv for Brotli support. Once you add that in this should work correctly.

```
if (req.http.Fastly-Orig-Accept-Encoding) {
    if (req.http.Fastly-Orig-Accept-Encoding ~ "\bbr\b") {
      set req.http.Accept-Encoding = "br";
    }
  }
```


MuriloDalRi (approved):

MuriloDalRi:
I think we also need to run `REGENERATE_EXPECTATIONS=1 bundle exec rspec` for the tests to pass

PR #337, opened 2021-05-19
Prevent modification of Fastly-Client-IP by client

bilbof:
As recommended in Fastly's documentation: https://developer.fastly.com/reference/http-headers/Fastly-Client-IP

If a client sets this header themselves, Fastly will use it. This has implications for us as we set the value of this header as the value of another header 'True-Client-IP', which is later set as the `real_ip_header` in [Nginx config at origin](https://github.com/alphagov/govuk-puppet/blob/32c1bbbb10067078c1406170666a135b4a10aaea/modules/router/templates/router_include.conf.erb#L12).

### DIFF
--- a/spec/test-outputs/www-integration.out.vcl
+++ b/spec/test-outputs/www-integration.out.vcl
@@ -121,6 +121,12 @@ sub vcl_recv {
 
   
 
+  # Protect header from modification at the edge of the Fastly network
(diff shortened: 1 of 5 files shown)


sihugh (approved):

bilbof:
Deployed to staging to exercise it a bit.

PR #341, opened 2021-06-15
Remove GCS configuration

ChrisBAshton:
Google Cloud Storage / Google Cloud Platform was deemed to be
removable in RFC 136, as its maintenance cost outweighs the
expected reliability benefit:
https://github.com/alphagov/govuk-rfcs/blob/main/rfc-136-remove-our-backup-cdn-in-gcp.md

This commit removes our unused GCS CDN configuration.


karlbaker02:
Just want to check something here, in that we're removing the config for the GCS mirror in this PR. Is this not separate from removing Google as our back-up CDN?

I don't want to re-open the already agreed RFC here, but it stands to reason that Fastly could be available and serving requests, whilst there are also issues affecting AWS such that we don't want to put all of our eggs in that AWS basket and instead fall back to the mirror on Google? Bruce asked something similar to this in the RFC, though not quite the same question. What are your thoughts @richardTowers ?

richardTowers (changes requested):
Yeah, Karl's right. RFC 136 was about the GCP CDN, not the files in the mirror. It even says:

> The mirroring approach works well. This RFC does not propose any change to the mirrors, including the Google Cloud Storage mirror.

I don't want to shut that conversation down - it may well be that having the additional mirror is overkill, but I think we need to discuss that separately.

A lot of the reasons the GCP CDN was a maintenance pain aren't relevant to the mirrors in GCS (e.g. we couldn't get a cert for the CDN, but we don't need a cert for Fastly to fall back to files in GCS).

ChrisBAshton:
My misunderstanding. Thanks for clarifying.

PR #342, opened 2021-06-21
Revert 310 add cookieless test

oscarwyatt:
Reverts #310 

Also makes minor refactor to reduce the amount of indenting


spec/fixtures/_multivariate_tests.vcl.erb.out
```diff
@@ -89,6 +89,4 @@ if (req.http.Cookie ~ "cookies_policy" && req.http.Cookie:cookies_policy ~ "%22u
     }
   }
 }
-else if (req.http.Cookie !~ "cookies_preferences_set") {
```
karlbaker02, line 92: Originally lines 2-3 of this file read:

```
if (req.http.Cookie ~ "cookies_policy") {
  if (req.http.Cookie:cookies_policy ~ "%22usage%22:true") {
```

You're not reverting that here, but this is fine. The reason I combined these lines into what is now line 2 of this current file:

```
if (req.http.Cookie ~ "cookies_policy" && req.http.Cookie:cookies_policy ~ "%22usage%22:true") {
```

is because I wanted to not need to repeat the checks of the previous lines 2-3 in my check here on line 92:

```
else if (req.http.Cookie !~ "cookies_preferences_set") {
```

(basically, if we get to the check on line 92, we can have confidence that either the cookies_policy cookie is not set, or the cookies_policy cookie is set but doesn't allow usage; if it had been left as it was before, the check on the cookies_policy cookie usage would have had to been repeated).

karlbaker02 (approved):
I have been through this, especially __multivariate_tests.vcl.erb_, with a fine-toothed comb, comparing to the additions made in https://github.com/alphagov/govuk-cdn-config/pull/310/files. All looks good, thanks for updating this revert PR - working in VCL with partials is not an easy feat 🥇  I would test this once again as you deploy across all environments just to be sure everything is still as expected.

PR #344, opened 2021-06-30
Add ExploreMenuAbTestable dictionaries in preparation for AB test

chao-xian:
## What

Add dictionaries for new AB Test. We don't want this active yet until we're all ready.

## Why

Supports the AB test on the menu header, where B group see Explore Team's Super Menu Header.

Trello card: https://trello.com/c/LUhrajXF/323-prepare-the-new-menu-a-b-test-for-sections-of-govuk


configs/dictionaries/exploremenuabtestable_percentages.yaml
```diff
@@ -0,0 +1,3 @@
+---
+  A: 100
+  B: 0
```
sihugh, line 3: should have a trailing new line

configs/dictionaries/ab_test_expiries.yaml
```diff
@@ -8,3 +8,4 @@
 Example: 86400
 # 91 days and 6 hours (~ 3 months)
 AccountBankHols: 7884000
+ExploreMenuAbTestable: 1209600
```
sihugh, line 11: if you're starting the AB test with 100 in A and then wanting to switch some people over, then having a long time here will make it harder to increase your B cohort quickly because they might all be stuck in A until their cookie expires. Could be worth starting with a shorter time, and then bumping it up once you have greater confidence? 

(It's up to you though, this isn't a problem as long as you're aware)

sihugh (changes requested):
Minor issue with your missing `/n` that might cause weird issues

configs/dictionaries/ab_test_expiries.yaml
chao-xian: Double checked with Tara and it's 1 day then 3 months. Updated in

PR #346, opened 2021-07-05
Add Start a Business intervention test

oscarwyatt:
Adds logic to show users the correct page based on how many SaB pages they've seen in their session

Have checked that this works as desired. The desirable behaviour is:

Should give users a cookie `ABTest-start_a_business_segment` which logs which segment they're in. 

When they first visit a SaB page (defined in the extra yaml file and assuming they accept additional tracking cookies):

* all users start in segment A 
* The rendering app is sent 2 headers, one defining whether the page is a SaB page, the other defining their segment (here: A)
* They are then given a cookie with a value of B, to indicate that they've visited one SaB page and that on the next SaB page request, they're in the B segment
* This cookie has an expiry time of 30 mins so that they are put back into the A segment after the session is expired
* All page requests where they already have the cookie will have the cookie reset with another 30 minute expiry time - so that the expiry time is re-started on every request

When users visit another SaB page with the B segment:

* they are shown the B variant (with the banner)
* The cookie is again reset on every page request with an additional 30 minutes

If the user visits a SaB page with `?hide-intervention=true`  query parameter (which is included in a link in the banner to hide the banner), or they visit the SaB checker (`/next-steps-for-your-business`)

* They are moved to segment C (which doesn't show the banner)
* The cookie has an expiry time of approx 6 months (significantly longer than the A/B period) so that they won't be shown the intervention even if they start a new session


vcl_templates/www.vcl.erb
```diff
@@ -374,6 +374,34 @@ sub vcl_recv {
     return(pass);
   }
 
+  if (req.http.Cookie ~ "cookies_policy") {
```
fredericfran-gds, line 377: might be better to put this `if` block one step above and group together all the cookies check?

vcl_templates/www.vcl.erb
```diff
@@ -536,6 +564,32 @@ sub vcl_deliver {
 <% end # if -%>
   # End dynamic section
 
+if (req.http.Cookie ~ "cookies_policy") {
```
fredericfran-gds, line 567: I don't know how sensible it is to "merge" this `if` statement with the previous one starting on line 551

fredericfran-gds (commented):
looks good to me, by no means an expert. I will let another reviewer give their opinion too.

vcl_templates/www.vcl.erb
oscarwyatt: Good shout :+1:

vcl_templates/www.vcl.erb
oscarwyatt: I thought about it, it felt a bit eggy because it means mixing up the logic for this with something else - I figured that this keeps the code a bit more separated which (I think) makes the commit history/PR a bit more legible. Happy to change it if you want me to though

fredericfran-gds: I prefer to keep a single tree of logic. Anyway, let's wait for Bill to give his comments.

karlbaker02:
So just to check:

Where you say
- _All page requests where they already have the cookie will have the cookie reset with another 30 minute expiry time - so that the expiry time is re-started on every request_
is when a user has a `B` variant?

Once a user is on the `C` variant, there is no way they will go back to an `A` or `B` variant for the next six months?

spec/test-outputs/www-integration.out.vcl
```diff
@@ -159,6 +159,34 @@ sub vcl_recv {
...
+      if (req.http.Cookie ~ "ABTest-start_a_business_segment=(A|B|C)") {
+        set var.sab_variant = re.group.1;
+        set req.http.GOVUK-ABTest-StartABusinessSegment = var.sab_variant;
+      }
+
+      if (table.lookup(start_a_business_ab_test_base_paths, querystring.remove(req.url)) == "true") {
+        # Header so rendering app will know it's a SaB page (and the config is in just one place)
+        set req.http.GOVUK-ABTest-IsStartABusinessPage = "true";
+        # Segment header
+        set req.http.GOVUK-ABTest-StartABusinessSegment = var.sab_variant;
```
karlbaker02, line 179: This is repeated work of line 172 but I assume you need this here for the case where the SaB cookie is not already set? It would probably be easier to understand if this is the case with an `else` block just to avoid any ambiguity of intention but appreciate this would create some duplicate code.

spec/test-outputs/www-integration.out.vcl
```diff
@@ -426,6 +454,32 @@ sub vcl_deliver {
   }
   # End dynamic section
 
+if (req.http.Cookie ~ "cookies_policy") {
+  if (req.http.Cookie:cookies_policy ~ "%22usage%22:true" && req.http.User-Agent !~ "^GOV\.UK Crawler Worker") {
```
karlbaker02, line 458: I'm curious both here and in the previous code in the receive block why we're not checking to ensure that the test is active as we do with the other tests? Ideally we should do this as a means to be able to control the test through configuration.

spec/test-outputs/www-integration.out.vcl
```diff
@@ -426,6 +454,32 @@ sub vcl_deliver {
...
+    declare local var.variant STRING;
+    set var.variant = req.http.GOVUK-ABTest-StartABusinessSegment;
```
karlbaker02, line 460: To check here, what will the value of `req.http.GOVUK-ABTest-StartABusinessSegment` return if it hasn't been set?

spec/test-outputs/www-integration.out.vcl
```diff
@@ -426,6 +454,32 @@ sub vcl_deliver {
...
+
+    declare local var.sab_expiry TIME;
+    if (var.variant == "C") {
+      # If they have opted out of the banner or have visited the SaB checker
+      # Expire in 26 weeks (approx 6 months)
+      set var.sab_expiry = time.add(now, std.integer2time(15724800));
```
karlbaker02, line 466: Is it also possible to use `time.add` as shown [in the docs](https://developer.fastly.com/reference/vcl/functions/date-and-time/time-add/) and specify the unit of time you want to add (avoiding a call to `std.integer2time()`?

spec/test-outputs/www-integration.out.vcl
```diff
@@ -426,6 +454,32 @@ sub vcl_deliver {
...
+    } else {
+      # Otherwise, expire in half an hour, approx this session
+      set var.sab_expiry = time.add(now, std.integer2time(1800));
+    }
+
+    if (req.http.GOVUK-ABTest-IsStartABusinessPage == "true" && var.variant == "A") {
```
karlbaker02, line 472: You probably want this to be an else if

spec/test-outputs/www-integration.out.vcl
```diff
@@ -426,6 +454,32 @@ sub vcl_deliver {
...
+      # If the user is in variant A and has seen a SaB page then we need to move them to B
+      # so they see the banner on the next SaB page they hit.
+      set var.variant = "B";
+    }
+
+    if (std.strlen(var.variant) > 0) {
```
karlbaker02, line 478: Will `var.variant` always be populated when we're running this test? If so, the check to ensure that the test is active in an earlier comment should be enough to be able to remove this check, because you're always setting the variant back in the block from line 162?

karlbaker02 (commented):
Good job 👍

spec/test-outputs/www-integration.out.vcl
oscarwyatt: Had to set up a fiddle to find out - it's an empty string

spec/test-outputs/www-integration.out.vcl
oscarwyatt: No, it's not guaranteed to be populated so we do want to keep the check - good question though

spec/test-outputs/www-integration.out.vcl
oscarwyatt: Do you mean something like this? 
```
    if (var.variant == "C") {
      # If they have opted out of the banner or have visited the SaB checker
      # Expire in 26 weeks (approx 6 months)
      set var.sab_expiry = time.add(now, 26w);
    } else if (req.http.GOVUK-ABTest-IsStartABusinessPage == "true" && var.variant == "A") {
      # Otherwise, expire in half an hour, approx this session
      set var.sab_expiry = time.add(now, 30m);
      # If the user is in variant A and has seen a SaB page then we need to move them to B
      # so they see the banner on the next SaB page they hit.
      set var.variant = "B";
    } else {
      # Otherwise, expire in half an hour, approx this session
      set var.sab_expiry = time.add(now, 30m);
    }
```

spec/test-outputs/www-integration.out.vcl
oscarwyatt: True, we can set all the values in the `start_a_business_ab_test_base_paths.yaml` to false which should have the same effect but happy to add it if it makes things a bit more standard

spec/test-outputs/www-integration.out.vcl
oscarwyatt: good shout :+1:

oscarwyatt:
> So just to check:
> 
> Where you say
> 
>     * _All page requests where they already have the cookie will have the cookie reset with another 30 minute expiry time - so that the expiry time is re-started on every request_
> 
> 
> is when a user has a `B` variant?
> 
> Once a user is on the `C` variant, there is no way they will go back to an `A` or `B` variant for the next six months?

Thanks for your comments, I _believe_ I've fixed those all now. To answer your questions:
yes, the A/B variant cookie expiry is 30 mins - so that if they start a new session they start afresh
Without deleting their cookies, no, there is no way back to A or B once they've been in C - we chose to do this because you can only end up in C if you choose to hide the banner or go to the actual checker that we're linking to - in which case it would be annoying to continue to see the banner. If we run another A/B within 6 months we'll have to use a slightly different cookie name

spec/test-outputs/www-integration.out.vcl
```diff
@@ -426,6 +459,43 @@ sub vcl_deliver {
...
+        set var.sab_expiry = time.add(now, 26w);
+      } else if (req.http.GOVUK-ABTest-IsStartABusinessPage == "true" && var.variant == "A") {
+        # Otherwise, expire in half an hour, approx this session
+        set var.sab_expiry = time.add(now, 30m);
+        # If the user is in variant A and has seen a SaB page then we need to move them to B
+        # so they see the banner on the next SaB page they hit.
+        set var.variant = "B";
+      } else {
+        # Otherwise, expire in half an hour, approx this session
+        set var.sab_expiry = time.add(now, 30m);
```
karlbaker02, line 490: If `var.variant` is empty do we want to be setting this expiry at all? (disregard if you're planning to keep the previous block, as it doesn't matter you're setting it if it's forcing the set-cookie response header. I was just thinking of ways to tidy it up (using something like `if (var.variant ~ "(A|B|C))"` and then a dictionary look-up on `var.variant` to remove all of the ifs completely).

spec/test-outputs/www-integration.out.vcl
```diff
@@ -426,6 +459,43 @@ sub vcl_deliver {
...
+
+      declare local var.sab_expiry TIME;
+      if (var.variant == "C") {
+        # If they have opted out of the banner or have visited the SaB checker
+        # Expire in 26 weeks (approx 6 months)
+        set var.sab_expiry = time.add(now, 26w);
+      } else {
+        # Otherwise, expire in half an hour, approx this session
+        set var.sab_expiry = time.add(now, 30m);
+      }
```
karlbaker02, line 476: Looks like an incomplete duplicate of lines 478-490 below (though with a good refactoring!).

spec/test-outputs/www-integration.out.vcl
karlbaker02: Cool. It'd be good to do that, not least if you need to turn it off for any reasons and have to update all of those values rather than a simple switch 🙂

spec/test-outputs/www-integration.out.vcl
karlbaker02: I see you've already updated this 👍

spec/test-outputs/www-integration.out.vcl
```diff
@@ -159,6 +159,39 @@ sub vcl_recv {
...
+        }
+
+        if (table.lookup(start_a_business_ab_test_base_paths, querystring.remove(req.url)) == "true") {
+          # Header so rendering app will know it's a SaB page (and the config is in just one place)
+          set req.http.GOVUK-ABTest-IsStartABusinessPage = "true";
+          # Segment header
+          set req.http.GOVUK-ABTest-StartABusinessSegment = var.sab_variant;
+        }
+
+        if (querystring.remove(req.url) ~ "next-steps-for-your-business" || std.strlen(querystring.get(req.url, "hide-intervention")) > 0) {
```
karlbaker02, line 186: Worth double checking that you may have both `GOVUK-ABTest-IsStartABusinessPage = "true"` set as well as going into this final `if` block - this is desired right? (assume it is)

spec/test-outputs/www-integration.out.vcl
oscarwyatt: whoops :facepalm:

spec/test-outputs/www-integration.out.vcl
oscarwyatt: Good shout - I think the new code is cleaner and yes we don't need to set this at all if the variant is empty :100:

spec/test-outputs/www-integration.out.vcl
oscarwyatt: Yes, this is desired, good question though

karlbaker02 (approved):
Looks good, great job 👍 💯

PR #356, opened 2021-09-02
Add custom GOVUK-Account-Session-Exists request header

barrucadu:
We will use this to show the appropriate navigation (signed in vs
signed out), and `Vary` based on it so we only cache one copy for all
signed in users and one copy for all signed out users, rather than one
copy each for all signed in users.

I don't think we need to add a case for:

    if (resp.http.Vary ~ "GOVUK-Account-Session-Exists")

because `~` is a regex match, so the `~ "GOVUK-Account-Session"` check
should be sufficient.

See also the progressive enhancement ADR for account-api[1].

[1] https://github.com/alphagov/account-api/blob/main/docs/adr/002-progressive-enhancement.md

---

[Trello card](https://trello.com/c/4wd8du6X/987-implement-navigation-caching-behaviour)

### DIFF
--- a/spec/test-outputs/www-integration.out.vcl
+++ b/spec/test-outputs/www-integration.out.vcl
@@ -154,9 +154,10 @@ sub vcl_recv {
 
 #FASTLY recv
 
-  # RFC 134
+  # GOV.UK accounts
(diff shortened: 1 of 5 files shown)


bevanloon (approved):
Looks okay to me.

PR #360, opened 2021-09-21
[DO NOT MERGE] Disable Explore Super Menu Header AB test

chao-xian:
We want to roll out the new header to all users, so we don't want to be AB testing any more.


chao-xian:
https://github.com/alphagov/govuk-cdn-config/pull/361 does a better job

PR #362, opened 2021-09-21
Use original client IP when checking ACLs

benthorner:
https://www.pivotaltracker.com/story/show/179659713

Previously we used "client.ip" to control access to Staging and
Integration, but this won't work for GOV.UK Alerts, since the IP
gets modified due to its use of the shielding feature [1]. This
moves setting the existing Fastly-Client-IP header earlier so that
we can reuse it as a reliable means of checking who's allowed in.

Disclaimer: I haven't tested this, since I don't have access to do
so. I'm fairly confident the header will always be set, so I don't 
expect any of the subsequent uses of it to fail.

[1]: https://developer.fastly.com/reference/vcl/variables/client-connection/client-ip/

### DIFF
--- a/spec/test-outputs/www-integration.out.vcl
+++ b/spec/test-outputs/www-integration.out.vcl
@@ -66,16 +66,21 @@ acl allowed_ip_addresses {
 
 
 sub vcl_recv {
+  # Protect header from modification at the edge of the Fastly network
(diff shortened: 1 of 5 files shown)


benthorner:
Closing as this isn't something we're actively working on.

PR #366, opened 2021-11-24
Remove the SaB intervention code

maxf:
This reverts [c936edfe164ef074f659d43d1294c98f33c31fe7](https://github.com/alphagov/govuk-cdn-config/commit/c936edfe164ef074f659d43d1294c98f33c31fe7)
The intervention banner experiment has now ended and this code is therefore no longer needed.


karlbaker02:
Thanks Max. Have you tested this out against integration CDN?

karlbaker02 (approved):
This looks like a clean revert, so as long as everything went fine with testing on integration then 👍

PR #369, opened 2022-03-02
Update Fastly purge request IP allow list

sihugh:
## Remove Fastly cache nodes from IP purge list

These IP addresses were [added in 2016](https://github.com/alphagov/govuk-cdn-config/commit/570d4e941f6c542b2dfcd96255b5c7b7916dee45) so that Fastly staff could purge our content if required, but only if they made the request from a Fastly cache node.

I'm not aware of this ever having been necessary since, and should it be so, we would be able to add the required IPs back again either here, or in the Fastly management console.

In addition, Fastly now advise [using API tokens to authenticate purge requests](https://docs.fastly.com/en/guides/authenticating-api-purge-requests).

I've added a [Trello card to the Platform Reliability team backlog][1] to switch this to using API tokens.

[1]: https://trello.com/c/O5Wy2X9w/2613-authenticate-fastly-purge-requests-with-an-api-token

## Remove Skyscape and Carrenza IPs

We no longer have infrastructure here that we would use to communicate with Fastly.

For context:
- Skyscape is UKCloud's former incarnation
- Carrenza is now owned by Six Degrees

Both were hosting providers for parts of GOV.UK in times of yore.

### DIFF
--- a/spec/test-outputs/www-integration.out.vcl
+++ b/spec/test-outputs/www-integration.out.vcl
@@ -37,24 +37,6 @@ acl purge_ip_allowlist {
   "34.248.44.175";    # AWS Integration NAT gateway
   "52.51.97.232";     # AWS Integration NAT gateway
 
-  "23.235.32.0"/20;   # Fastly cache node
-  "43.249.72.0"/22;   # Fastly cache node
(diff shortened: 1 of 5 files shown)


sengi (approved):
LGTM - thanks for cleaning this up!

PR #372, opened 2022-03-16
Add symlinks for www-eks and assets-eks services.

sengi:
This is probably the simplest way to allow rolling out the `*-eks` configs from the `CDN: deploy service` Jenkins job. We need to use unique service names (rather than additional environments within the existing services) for the EKS configs so that they can be chosen in the "Build with parameters" page on Jenkins.

Alternatively we could modify the script to special-case the `*-eks` configs, but that seems even bodgier than just adding these symlinks. Once GOV.UK traffic is permanently being served from Kubernetes, we'll delete these extra configs anyway.

See alphagov/govuk-cdn-config-secrets#149.

#### Testing

Deployed locally with `bundle exec deploy_service www-eks integration`.

#### Tips for reviewing

Best reviewed commit-by-commit, since a lot of the change is just auto-generated golden test outputs in commit deb712e. The interesting bit is [`ba653eb` Add symlinks for www-eks and assets-eks services](https://github.com/alphagov/govuk-cdn-config/pull/372/commits/ba653eb5ba73e530af7ec767dce2658918e44cba).

[Trello](https://trello.com/c/9BPpFRBe/851)

### DIFF
--- a/lib/deploy_dictionaries.rb
+++ b/lib/deploy_dictionaries.rb
@@ -76,7 +76,7 @@ class DeployDictionaries
                   rescue StandardError
                     nil
                   end
-    raise "Unknown vhost/environment combination" unless config_hash
+    raise "Unknown vhost/environment combination: #{vhost} #{environment}" unless config_hash
 
     config_hash
   end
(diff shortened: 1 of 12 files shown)


sengi:
Urgh, need to fix those awful change detector tests. Stand by.

fredericfran-gds (approved):
lgtm

PR #373, opened 2022-03-21
Strip cookies for applications that don't them at fastly

samsimpson1:
This functionality is currently handled by varnish in EC2, but as part of the move to EKS we need to move this to fastly.

Trello: https://trello.com/c/HXtsiNCO/855-strip-cookies-at-fastly

### DIFF
--- a/spec/test-outputs/www-eks-integration.out.vcl
+++ b/spec/test-outputs/www-eks-integration.out.vcl
@@ -124,6 +124,18 @@ sub vcl_recv {
   }
 
 
+  # Strip cookies from inbound requests. Corresponding rule in vcl_fetch{}
+  # For simplicity and security most applications should not use cookies.
+  # With the exception of:
+  #   - Licensing
+  #   - email-alert-frontend (for subscription management)
+  #   - frontend (for sessions across funding registration form)
+  #   - smart answers coronavirus find support flow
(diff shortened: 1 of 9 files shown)


sengi (approved):
I'm not very familiar with VCL but from what I've been able to look up, I think this should be fine.

Line 467 enables caching of responses which have a `set-cookie` header; this normally would ring alarm bells given that we're not stripping *all* `set-cookie` response headers, but it appears we have the same arrangement [in govuk-puppet] and there's a comment there explaining that the apps are supposed to set appropriate `cache-control` headers, so fair enough.

[in govuk-puppet]: https://github.com/alphagov/govuk-puppet/blob/6b8b8be/modules/varnish/templates/default.vcl.erb#L169

sengi:
I presume you're planning on rolling this out to staging and prod once you've tested it in integration? Also how do you plan to test it once it's deployed to integration?

samsimpson1:
> I presume you're planning on rolling this out to staging and prod once you've tested it in integration? Also how do you plan to test it once it's deployed to integration?

To test, we will try the services that require cookies and ensure they still function. For the ones that don't, we will capture some traffic and examine the HTTP headers to ensure cookies aren't being sent to the applications from fastly

sengi:
SGTM 👍
cookie monster

PR #376, opened 2022-03-22
Remove the feature to set header in the backend

fredericfran-gds:
This feature was added because the ingress in EKS was not configured
properly to accept the original host header of fastly.

This had now been fixed in [here](https://github.com/alphagov/govuk-helm-charts/pull/270)


sengi (approved):
Cheers for tidying this up

PR #379, opened 2022-05-16
Copy Accept-Encoding normalisation from EC2 Varnish

samsimpson1:
This is copied from the EC2-based Varnish config. This is required for replatforming work, since we will be retiring the EC2-based Varnish services.

I have tested it using Fastly Fiddle.

This is integration-only at this point to test before deploying everywhere.

Trello: https://trello.com/c/88bU7Qmm/854-normalise-accept-encoding-header-at-fastly

### DIFF
--- a/vcl_templates/www.vcl.erb
+++ b/vcl_templates/www.vcl.erb
@@ -204,6 +204,15 @@ sub vcl_recv {
   }
   <% end %>
 
+  <% if environment == 'integration' %>
+  # Strip Accept-Encoding header if the content is already compressed
+  if (req.http.Accept-Encoding) {
+    if (req.url ~ "\.(jpeg|jpg|png|gif|gz|tgz|bz2|tbz|zip|flv|pdf|mp3|ogg)$") {
+      remove req.http.Accept-Encoding;
+    }
+  }
+  <% end %>
+
   # Force SSL.
   if (!req.http.Fastly-SSL) {
      error 801 "Force SSL";
--- a/spec/test-outputs/www-eks-integration.out.vcl
+++ b/spec/test-outputs/www-eks-integration.out.vcl
@@ -58,6 +58,15 @@ sub vcl_recv {
 
   
 
+  
+  # Strip Accept-Encoding header if the content is already compressed
+  if (req.http.Accept-Encoding) {
+    if (req.url ~ "\.(jpeg|jpg|png|gif|gz|tgz|bz2|tbz|zip|flv|pdf|mp3|ogg)$") {
+      remove req.http.Accept-Encoding;
+    }
+  }
+  
+
   # Force SSL.
   if (!req.http.Fastly-SSL) {
      error 801 "Force SSL";
--- a/spec/test-outputs/www-eks-production.out.vcl
+++ b/spec/test-outputs/www-eks-production.out.vcl
@@ -151,6 +151,8 @@ sub vcl_recv {
 
   
 
+  
+
   # Force SSL.
   if (!req.http.Fastly-SSL) {
      error 801 "Force SSL";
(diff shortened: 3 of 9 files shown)


vcl_templates/www.vcl.erb
```diff
@@ -202,6 +202,22 @@ sub vcl_recv {
   }
   <% end %>
 
+  # Normalise Accept-Encoding header
```
sengi, line 205: Fastly already does some similar-looking stuff as standard - have you checked that this isn't replicating the snippet that Fastly adds itself? (if you view the full VCL in the Fastly UI, you should see an extra stanza or two of VCL that isn't coming from our templates - I think it's in there).

https://developer.fastly.com/reference/http/http-headers/Accept-Encoding/#normalization

samsimpson1: Ah, I wasn't aware of that. We can keep the bit that strips the header if the content is already compressed, but I don't know if that's worth bothering with

sengi: Ah, sorry - I was partly mistaken - the [config I was thinking of](https://github.com/alphagov/govuk-cdn-config/blob/f95e88d/vcl_templates/www.vcl.erb#L384) is actually ours.

Leads me to another question though, which is: have you tested that this doesn't break Brotli?

sengi: I still have a hunch that we're duplicating Fastly functionality but I'm less sure. It's important that we don't break Brotli though, since that would be a regression.

samsimpson1: Sorry yes, this does seem to be replicating what fastly is already doing. I copied this directly from the EC2 Varnish VCL file, so we can definitely get rid of the bits that check for gzip and deflate. We can probably keep the bit that checks if the content is already compressed, though.

The config that I copied and that which Fastly already does shouldn't work with brotli at all, but it shouldn't break anything. We're manipulating a request header here before it gets to a backend server that performs the compression, so if a browser says it accepts brotli, the server will never see that and will only serve an uncompressed file or one done with gzip/deflate

sengi: Oh, are you saying the [Brotli support](https://github.com/alphagov/govuk-cdn-config/blob/f95e88d/vcl_templates/www.vcl.erb#L383) has never worked? 😅

samsimpson1: As far as I can tell, that should never have worked because an `Accept-Encoding: br` header would be stripped out in EC2's Varnish: https://github.com/alphagov/govuk-puppet/blob/56c3da336aadb68d7dfbf6b0db2e748b18221bbc/modules/varnish/templates/default.vcl.erb#L49-L63

samsimpson1: I've updated this PR to only remove the Accept-Encoding header if the content is already compressed. The functionality Fastly provides should do the rest.

sengi: I'm pretty sure Brotli worked before. https://trello.com/c/u9qiWQr1/13-3-enable-brotli-compression-in-fastly-for-production

Bear in mind that it's Fastly that serves the Brotli-compressed objects, not us. So as long as we're not breaking the header in vcl_recv (i.e. the procedure that runs before Fastly processes the incoming request) we should be ok. Please do test that we haven't broken it though.

sengi: Actually now I'm not sure that I'm not just getting confused here. Anyway, please test for regressions! Code looks good to me.

sengi (approved):
LGTM (chatted out of band about it being integration-only - that's fine if you wanna do separate PRs to roll it out the rest of the way)

PR #377, opened 2022-03-28
Add AB test for New Browse pages 

chao-xian:
This is an AB Test for redesigned Mainstream Browse pages rendered by Collections. See [corresponding PR](https://github.com/alphagov/collections/pull/2711)

[Trello](https://trello.com/c/ACFzuepy/850-prep-prs-for-ab-test-of-browse-pages)

PR #380, opened 2022-05-16
Revert "Add AB test for New Browse pages "

chao-xian:
Reverts alphagov/govuk-cdn-config#377


hannako:
preceded by other work.

PR #385, opened 2022-05-17
Have the deploy scripts run a shallow git clone.

sengi:
We only need to fetch the head of the main branch of `govuk-cdn-config-secrets`. This speeds up deployment and saves a little bit on resource usage.

Tested: scripts still work locally.


nsabri1 (approved):
LGTM

PR #391, opened 2022-05-18
Don't strip cookies in EC2-backed services, only EKS ones.

sengi:
This was breaking the A/B test smoke tests in EC2 integration, which causes confusion. I'm not sure we even need to do cookie stripping in the EC2-backed services.

Partially reverts #373.

https://trello.com/c/HXtsiNCO


nsabri1 (approved):
LGTM

PR #396, opened 2022-07-01
Enable IP allowlist for the EKS production environment

theseanything:
Whilst we are testing deploying the production environment we'd like restrict access from the general public.

### DIFF
--- a/vcl_templates/www.vcl.erb
+++ b/vcl_templates/www.vcl.erb
@@ -40,7 +40,7 @@ backend F_origin {
 <% end -%>
 }
 
-<% if %w(staging production).include?(environment) %>
+<% if %w(staging production production-eks).include?(environment) %>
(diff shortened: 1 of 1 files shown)


richardTowers (commented):
Do we not need

```
<% if not %w(production).include?(environment) %>
```

on L166 updating too?

theseanything:
> Do we not need
> 
> ```
> <% if not %w(production).include?(environment) %>
> ```
> 
> on L166 updating too?

I don't think so - we're saying _if the environment is NOT production then include the allowlists_. Which will now be true for production in EKS as `production-eks != production`

richardTowers:
Oh yeah 🤦🏻. I can't do boolean logic on Fridays.

richardTowers (approved):

PR #397, opened 2022-07-08
Strip Accept-Encoding header in all environments

samsimpson1:
Tests in integration look good, should be safe to roll out everywhere.

See also: https://github.com/alphagov/govuk-cdn-config/pull/379

https://trello.com/c/HXtsiNCO/855-strip-cookies-at-fastly

### DIFF
--- a/spec/test-outputs/www-eks-integration.out.vcl
+++ b/spec/test-outputs/www-eks-integration.out.vcl
@@ -58,14 +58,12 @@ sub vcl_recv {
 
   
 
-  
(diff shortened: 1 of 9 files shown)


theseanything:
@samsimpson1 would be nice to include the reasoning as to why we strip this heading in the commit message for future reference.

samsimpson1:
@theseanything I've amended the commit message to be more descriptive

sengi (approved):

PR #399, opened 2022-07-18
Remove NewBrowse AB Test

chao-xian:
[Trello](https://trello.com/c/glBoTClk/1158-switch-off-new-browse-ab-test-and-show-new-browse-instead-of-miller-columns)

We've gathered enough data and need a small break between this ending and the next AB Test that we want to set up.


AgaDufrat (approved):
LGTM

hannako (approved):
👍

PR #405, opened 2022-08-19
Add config for replatforming traffic split experiment

samsimpson1:
This is a temporary change to enable sending a proportion of traffic to the new k8s GOV.UK stack. It is enabled with a config variable, and has been set up for use in staging before the production test next week.

https://trello.com/c/a6qhzsAy/987-create-fastly-config-for-production-traffic-split-experiment

### DIFF
--- a/vcl_templates/www.vcl.erb
+++ b/vcl_templates/www.vcl.erb
@@ -40,6 +40,55 @@ backend F_origin {
 <% end -%>
 }
 
+<%- if config['replatforming_traffic_split'] %>
+backend F_eks_origin {
+    .connect_timeout = 5s;
+    .dynamic = true;
+    .port = "443";
+    .host = "www-origin.eks.staging.govuk.digital";
+    .first_byte_timeout = 15s;
+    .max_connections = 200;
+    .between_bytes_timeout = 10s;
+    .share_key = "<%= config.fetch('service_id') %>";
+
+    .ssl = true;
+    .ssl_check_cert = <%= config['disable_tls_validation'] ? 'never' : 'always' %>;
+    .min_tls_version = "<%= config.fetch('min_tls_version', '1.2') %>";
+    <%- if config['ssl_ciphers'] -%>
+    .ssl_ciphers = "<%= config['ssl_ciphers'] -%>";
+    <%- end -%>
+    .ssl_cert_hostname = "www-origin.eks.staging.govuk.digital";
+    .ssl_sni_hostname = "www-origin.eks.staging.govuk.digital";
+
+<% if config['probe'] -%>
+    .probe = {
+        .request =
+            "HEAD <%= config['probe'] %> HTTP/1.1"
+            "Host: <%= config.fetch('origin_hostname') %>"
+            "User-Agent: Fastly healthcheck (git version: <%= config['git_version'] %>)"
+<% if config['rate_limit_token'] -%>
+            "Rate-Limit-Token: <%= config['rate_limit_token'] %>"
+<% end -%>
+<% if config['basic_authentication'] -%>
+            "Authorization: Basic <%= config['basic_authentication'] %>"
+<% end -%>
+            "Connection: close";
+        .threshold = 1;
+        .window = 2;
+        .timeout = 5s;
+        .initial = 1;
+        .expected_response = 200;
+        .interval = 10s;
+    }
+<% end -%>
+}
+
+director D_traffic_split random {
+  { .backend = F_origin; .weight = 1; }
+  { .backend = F_eks_origin; .weight = 99; }
+}
+<%- end %>
+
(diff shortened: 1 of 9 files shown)


vcl_templates/www.vcl.erb
```diff
@@ -40,6 +40,55 @@ backend F_origin {
 <% end -%>
 }
 
+<%- if config['replatforming_traffic_split'] %>
+backend F_eks_origin {
+    .connect_timeout = 5s;
+    .dynamic = true;
+    .port = "443";
+    .host = "www-origin.eks.staging.govuk.digital";
```
fredericfran-gds, line 48: template things out now.

vcl_templates/www.vcl.erb
```diff
@@ -40,6 +40,55 @@ backend F_origin {
...
+    .between_bytes_timeout = 10s;
+    .share_key = "<%= config.fetch('service_id') %>";
+
+    .ssl = true;
+    .ssl_check_cert = <%= config['disable_tls_validation'] ? 'never' : 'always' %>;
+    .min_tls_version = "<%= config.fetch('min_tls_version', '1.2') %>";
+    <%- if config['ssl_ciphers'] -%>
+    .ssl_ciphers = "<%= config['ssl_ciphers'] -%>";
+    <%- end -%>
+    .ssl_cert_hostname = "www-origin.eks.staging.govuk.digital";
```
fredericfran-gds, line 60: template things out now

vcl_templates/www.vcl.erb
```diff
@@ -40,6 +40,55 @@ backend F_origin {
...
+<% end -%>
+            "Connection: close";
+        .threshold = 1;
+        .window = 2;
+        .timeout = 5s;
+        .initial = 1;
+        .expected_response = 200;
+        .interval = 10s;
+    }
+<% end -%>
```
fredericfran-gds, line 83: Maybe this end should be on the next line, have you tried a manual general of the template where this is activated?

vcl_templates/www.vcl.erb
```diff
@@ -40,6 +40,55 @@ backend F_origin {
 <% end -%>
 }
 
+<%- if config['replatforming_traffic_split'] %>
+backend F_eks_origin {
```
fredericfran-gds, line 44: We may have an issue with the number of backends (5 is the default limit if I remember correctly)  but we shall see. Ref: https://docs.fastly.com/en/guides/resource-limits#service-domain-and-origin-limits

fredericfran-gds (changes requested):
Great work. 

See in line comments.

In addition, maybe adding some more logic on line 379 is needed.
<% if config['basic_authentication'] -%>
  if (req.backend == F_origin ) {
    set req.http.Authorization = "Basic <%= config['basic_authentication'] %>";
  }
<% end -%>

vcl_templates/www.vcl.erb
samsimpson1: I didn't want to template because it's only going to be used for this experiment. Once I've verified it works in staging, this will be rolled back and I'll put in a new PR with the actual production values.

vcl_templates/www.vcl.erb
samsimpson1: I think this should bring us up to the 5 backend limit (as long as directors don't count as backends, which they shouldn't)

vcl_templates/www.vcl.erb
samsimpson1: That entire block was copied from here: https://github.com/alphagov/govuk-cdn-config/blob/0368e27e7b7bbcbd3c31766a55e88dd81b32d404/vcl_templates/www.vcl.erb#L20-L40

It looked good when I generated the template locally too.

samsimpson1:
> In addition, maybe adding some more logic on line 379 is needed. <% if config['basic_authentication'] -%> if (req.backend == F_origin ) { set req.http.Authorization = "Basic <%= config['basic_authentication'] %>"; } <% end -%>

Basic auth isn't used in staging/production, so that statement will never pass.

vcl_templates/www.vcl.erb
sengi: Can we safely test this on the prod service on Tuesday (tomorrow) perhaps? Or are we confident that staging has the same number of existing backends so we're ok?

vcl_templates/www.vcl.erb
sengi: I'm ok with either approach (as long as you revalidate the solution in staging if you change the approach). If we want to change the approach, it'd be good to be specific about the practical reasons behind that as applicable to our use case.

sengi (approved):

fredericfran-gds (approved):
Fine, I wanted it to be generic rather than for staging only but if Chris is fine with it.

PR #406, opened 2022-08-22
Update replatforming traffic split URLs to production; set traffic split ratio to 5%

samsimpson1:
Test in staging was successful; this sets the backend URLs to the production ones, and sets the ratio to 1:20 (5%) going to the EKS platform.

https://trello.com/c/a6qhzsAy/987-create-fastly-config-for-production-traffic-split-experiment


fredericfran-gds (approved):
I would have much preferred that the eks_origin is templated but oh well.

PR #408, opened 2022-08-25
Template eks origin hostname; use sticky sessions for replatforming traffic split

samsimpson1:


sengi:
For anyone following along, strictly speaking this isn't the scary kind of sticky sessions; just consistent hash by source IP.

sengi (approved):

PR #413, opened 2022-10-05
Set X-Forwarded-Host headers

theseanything:
This ensures that Fastly sets the X-Forwarded-Host headers to a single value that is the same original request host. This helps mitigates against clients for spoofing the X-Forwarded-Host header and prevents cache poisoning. This must be set in the vcl_miss and vcl_pass subroutines as they occur before backend requests and have access to the the `bereq` token.

This is suggested from the Fastly docs here:
https://developer.fastly.com/reference/http/http-headers/X-Forwarded-Host/ 

This assumes that backend requests are only for client requests that have valid and non-spoofed hosts, which is fine as Fastly will only be dealing with request that have to correct Host header for the service.

e.g. curl -H 'Host: phishing.com' https://assets.publishing.service.gov.uk/ won't be handled by our Fastly service, as Fastly will look for the "phishing.com" service, not the "assets.publishing.service.gov.uk" service.

### DIFF
--- a/spec/test-outputs/assets-eks-integration.out.vcl
+++ b/spec/test-outputs/assets-eks-integration.out.vcl
@@ -143,6 +143,11 @@ sub vcl_miss {
 #FASTLY miss
 }
 
+sub vcl_pass {
+  set bereq.http.X-Forwarded-Host = req.http.host;
+#FASTLY pass
+}
+
 sub vcl_deliver {
 #FASTLY deliver
 }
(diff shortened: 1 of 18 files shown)


theseanything:
Closed in favour of https://github.com/alphagov/govuk-cdn-config/pull/414. Moved the request headers to the vcl_recv subroutine instead of vcl_miss and vcl_pass.

PR #415, opened 2022-10-05
More flexible template rendering with partials

kevindew:

This makes a couple of changes to the RenderTemplate class:
- It now accepts an argument of locals which sets local variables, this allows the RenderTemplate class to be used for bouncer VCL templates
- It adds a `#render_partial` to allow the embedding of partial files inside a VCL template

My motivation for adding this in is that I want to add the same partial to a number of VCL templates and was finding myself bogged down with repetitive code and implementation inconsistencies. This work is applied in: https://github.com/alphagov/govuk-cdn-config/pull/416 which is currently open as a draft.

This is best reviewed commit by commit as most of the diff is just churn from auto generated files (some having indentation fixed others added that were missing before).

### Rendering Template

Prior to this change you would render a template with:

```ruby
RenderTemplate.render_template("template", environment, config, version)
```

Now you do:

```ruby
RenderTemplate.call("template", locals: { environment: environment, config: config, version: version })
```

### Rendering a partial

To render a partial you previously did:

```erb
  <% template_path = File.join(File.dirname(__FILE__), "vcl_templates/_multivariate_tests.vcl.erb") -%>
  <%= ERB.new(File.new(template_path).read, nil, "-", "_erbout2").result(binding) %>
```

Now it's:

```erb
<%= render_partial("multivariate_tests", locals: { ab_tests: ab_tests }) %>
```

### Indenting a partial

You can pass an indentation argument to a partial and the VCL will be indented:

```erb
<%= render_partial("my_partial", indentation: "    ") %>
```
can produce:

```
    if (req.request != "HEAD" && req.request != "GET" && req.request != "FASTLYPURGE") {
      return(pass);
    }
```

### DIFF
--- a/lib/render_template.rb
+++ b/lib/render_template.rb
@@ -1,8 +1,20 @@
 class RenderTemplate
-  def self.render_template(configuration, environment, config, version)
-    # Both config and ab_tests are used inside the vcl.erb template
-    vcl_file = File.join(File.dirname(__FILE__), "..", "vcl_templates", "#{configuration}.vcl.erb")
-    ab_tests = YAML.load_file(File.join(__dir__, "..", "ab_tests", "ab_tests.yaml"))
-    ERB.new(File.read(vcl_file), trim_mode: "-").result(binding)
+  attr_reader :template, :locals
+
+  def self.call(...)
+    new(...).call
+  end
+
+  def initialize(template, locals: {})
+    @template = template
+    @locals = locals
+  end
+
+  def call
+    bind = binding
+    locals.each { |k, v| bind.local_variable_set(k, v) }
+
+    embedded_ruby = File.read("#{__dir__}/../vcl_templates/#{template}.vcl.erb")
+    ERB.new(embedded_ruby, trim_mode: "-").result(bind)
   end
 end
--- a/lib/render_template.rb
+++ b/lib/render_template.rb
@@ -11,10 +11,36 @@ class RenderTemplate
   end
 
   def call
-    bind = binding
-    locals.each { |k, v| bind.local_variable_set(k, v) }
-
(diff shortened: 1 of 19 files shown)


lib/render_template.rb
```diff
@@ -11,10 +11,32 @@ def initialize(template, locals: {})
...
     vcl_file = File.join(File.dirname(__FILE__), "..", "vcl_templates", "#{template}.vcl.erb")
-    ERB.new(File.read(vcl_file), trim_mode: "-").result(bind)
+    ERB.new(File.read(vcl_file), trim_mode: "-").result(binding_with_local_variables(locals))
+  end
+
+  def render_partial(partial, indentation: "", locals: {})
+    partial_path = File.join(File.dirname(__FILE__), "..", "vcl_templates", "_#{partial}.vcl.erb")
+
+    output = ERB
+      .new(File.read(partial_path), trim_mode: "-", eoutvar: "_#{partial}_eoutvar")
```
sihugh, line 22: I find this line (and the similar one elsewhere) is quite hard to read. Is it worth splitting it up?

spec/test-outputs/www-test.out.vcl
```diff
@@ -144,87 +144,86 @@ sub vcl_recv {
...
+        } else {
+          set req.http.GOVUK-ABTest-LevelTwoBrowse = "Z";
+        }
       }
     }
   }
-}
-# End dynamic section
-
+  # End dynamic section
```
sihugh, line 227: These diffs are 👨‍🍳 💋

sihugh (approved):
Nitpick about maybe splitting up a long statement, but LGTM. ✅

sihugh (approved):

PR #416, opened 2022-10-05
Consistently set headers in Fastly as public boundary

kevindew:

This is a draft that requires: https://github.com/alphagov/govuk-cdn-config/pull/415 and will need rebasing after: https://github.com/alphagov/govuk-cdn-config/pull/414 is merged. I've opened it up to show the intention of boundary headers as a) motivation for #415 and b) to get feedback on what is set.

It's easiest to look at the first commit https://github.com/alphagov/govuk-cdn-config/commit/ca109d332638435d81995fa5b1357c8dbceb0645 to get a gist of this PR as the second commit is a lot of test output churn.

---

This applies the same header setting approach for requests at the Fastly
boundary of all our Fastly services that proxy.

The motivation for this is to be clearer in what we trust header wise
and not risk situations where our applications can behave differently
based on headers a user sets (cache poisoning).

This moves the same client-IP header code that was in most of the files
to one place. This sets a header of True-Client-IP based on the IP that
connected to Fastly. We also remove the Client-IP header, I'm continuing
to do that based on the precedence but I couldn't identify what that
precedent was.

We set the X-Forwarded-For header to be the IP that communicated with
Fastly, prior to a request reaching Fastly we can't trust that the data
in here isn't a user specifying a fake IP address.

We reset the X-Forwarded-Host to be the host that Fastly received a
request for, this prevents a user pretending that the request originated
from a different host. By default Fastly will append to this header, as
Fastly doesn't know if the request to it is from a trusted source or
not.

We set X-Forwarded-Server to the Fastly server hostname. I'm not sure if
this is actually of value, but it seems better to override it than ever
be caught out by a user setting this and we think it's from within our
network.

### DIFF
--- a/vcl_templates/_boundary_headers.vcl.erb
+++ b/vcl_templates/_boundary_headers.vcl.erb
@@ -1,0 +1,16 @@
+# Protect header from modification at the edge of the Fastly network
+# https://developer.fastly.com/reference/http-headers/Fastly-Client-IP
+if (fastly.ff.visits_this_service == 0 && req.restarts == 0) {
+  set req.http.Fastly-Client-IP = client.ip;
+}
+
+# Original client address (e.g. for rate limiting).
+set req.http.True-Client-IP = req.http.Fastly-Client-IP;
+
+# Reset proxy headers at the boundary to our network so we can trust them in our stack
+set req.http.X-Forwarded-For = req.http.Fastly-Client-IP;
+set req.http.X-Forwarded-Host = req.http.host;
+set req.http.X-Forwarded-Server = server.hostname;
+
+# Discard user specified headers that we don't want to trust
+unset req.http.Client-IP;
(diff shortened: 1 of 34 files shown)


vcl_templates/_boundary_headers.vcl.erb
```diff
@@ -0,0 +1,16 @@
+# Protect header from modification at the edge of the Fastly network
+# https://developer.fastly.com/reference/http-headers/Fastly-Client-IP
+if (fastly.ff.visits_this_service == 0 && req.restarts == 0) {
+  set req.http.Fastly-Client-IP = client.ip;
+}
+
+# Original client address (e.g. for rate limiting).
+set req.http.True-Client-IP = req.http.Fastly-Client-IP;
```
sihugh, line 8: I'm considering tweaking this on staging in future so that we (if we [include original IPs in goreplay requests](https://github.com/buger/goreplay/wiki/Capturing-and-replaying-traffic#tracking-original-ip-addresses)) we can more effectively test changes to our rate limiting policies

(But this looks fine)

sihugh (approved):
This all looks good to me, thanks 👍

ChrisBAshton (approved):
⭐

PR #418, opened 2022-10-06
Set proxy headers with original request headers

theseanything:
This fixes an issue where the X-Forwarded-Host header was incorrect, as it was being set from the request Host header which was being re-written previously. This moves the logic to set the proxy headers earlier in the vcl_recv subroutine as to ensure they are using the original header information.

Been tested in integration - fixed the X-Forwarded-Host header to actually be set as the request Host.


kevindew (approved):
d'oh! that pesky host override.

PR #419, opened 2022-10-07
Fix IP allow lists for Production EKS

theseanything:
We previously renamed the Production EKS service/environment from production-eks to production. However forgot to update conditional to apply IP allow lists only to production-eks.


sengi (approved):
Awesome! Thanks for fixing!

PR #424, opened 2022-11-25
Enable traffic split for asset services

theseanything:
This adds config to enable split traffic for the asset services to send traffic to the EKS cluster. This is to be used for our live traffic experiments.


samsimpson1 (approved):
LGTM

PR #428, opened 2022-11-29
Set traffic split for www and assets to 20%

theseanything:


sengi (approved):

nsabri1 (approved):
OK

PR #433, opened 2022-12-01
Use healthcheck hostname for probes instead of origin hostname for EKS backend

samsimpson1:

### DIFF
--- a/vcl_templates/www.vcl.erb
+++ b/vcl_templates/www.vcl.erb
@@ -64,7 +64,7 @@ backend F_eks_origin {
     .probe = {
         .request =
             "HEAD <%= config['probe'] %> HTTP/1.1"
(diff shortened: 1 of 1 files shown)


theseanything (approved):

sengi (approved):

vcl_templates/www.vcl.erb
```diff
@@ -64,7 +64,7 @@ backend F_eks_origin {
     .probe = {
         .request =
             "HEAD <%= config['probe'] %> HTTP/1.1"
-            "Host: <%= config.fetch('origin_hostname') %>"
+            "Host: <%= config.fetch('healthcheck_hostname', config['eks_origin_hostname']) %>"
```
theseanything, line 67: We should probably not default back to the origin header, as it removes the safety of the health check.

PR #434, opened 2022-12-02
Health check probe updates for traffic split

samsimpson1:


sengi (approved):
Have you verified exhaustively that every instantiation of this template sets `healthcheck_hostname`? Otherwise LGTM 👍

samsimpson1:
> Have you verified exhaustively that every instantiation of this template sets `healthcheck_hostname`?

I've tested all of the www/assets hosts in every environment and they all respond as I'd expect

PR #441, opened 2023-02-02
Strip cookies after using setting up A/B test

theseanything:
This moves the logic to strip cookie after the logic for setting up the A/B test in the recv vcl routine. This is because the A/B tests need access to the cookie to understand whether a user has already been assigned a variant.


sengi (approved):
Aha good find!

PR #448, opened 2023-03-19
Work around Ruby 3.1 gratuitous YAML API breakage.

sengi:
Looks like `deploy_dictionaries` wasn't tested after upgrading to Ruby 3.1. Sigh.


MuriloDalRi (approved):

sengi:
Cheers!

PR #450, opened 2023-03-20
Set traffic split ratio to 1:1

samsimpson1:


nsabri1 (approved):
ok

PR #453, opened 2023-04-18
Strip query params from /alerts URLs

sihugh:
There is no requirement for query params to reach origin on the /alerts paths and we don't require them for cache busting or other features within Fastly.

By dropping these, we improve our cache hit rate at the edge and our ability to cater for large amounts of traffic.

This regex should match:

/alerts
/alerts/
/alerts/something
/alerts/something/longer

but not

/random/alerts
/alerts-go-missing
/alerts-go-missing/example

We're matching on the path, so we shouldn't need to worry about params here.


vcl_templates/www.vcl.erb
```diff
@@ -303,6 +303,11 @@ sub vcl_recv {
     set req.url = querystring.remove(req.url);
   }
 
+  if (req.url.path ~ "^\/alerts(\/.*)?$") {
```
sengi, line 306: simplifies to
```suggestion
  if (req.url.path ~ "^\/alerts\/?") {
```

(I double-checked that VCL's `~` is indeed a [partial match](https://docs.fastly.com/en/guides/vcl-regular-expression-cheat-sheet#matching-at-the-beginning-or-end-of-a-string) rather than a full match.)

sengi (approved):

vcl_templates/www.vcl.erb
sihugh: Lovely stuff, thanks 👍

sengi: (updated the suggestion)

It's not quite as much of a simplification as I'd first thought, but it does at least save some computation and possibly some memory 🤷 

Some quick tests to check I've not messed up: https://regexr.com/7ca86

sengi (approved):

PR #454, opened 2023-05-03
Uxm ab test

Rosa-Fox:
Practising setting up a AB test. [Corresponding rendering application code here](https://github.com/alphagov/collections/pull/3229/files).


configs/dictionaries/uxmpracticetest_percentages.yaml
```diff
@@ -0,0 +1,4 @@
+---
+A: 50
```
sihugh, line 2: If you're doing this for a real test, I tend to ship it with the current version set to 100% and any variants set to 0. (It also helps to set a short cookie time while doing this bit)
This approach gives you a bit of time to check that cookies are set correctly and just kick the tyres. You can test the other variants with the GOV.UK chrome extension at this point.

Once you're confident everything is OK, you can start upping the variant percentages and the cookie age.

It's probably also important to know that because the test adds a `Vary` header for your test, it invalidates Fastly caches for the affected pages when that header starts to appear. That's probably not super important for this page at the moment, but if you happen to be doing a test on a mega-popular page that has a lot of backend work, it can be a factor.

ab_tests/ab_tests.yaml
```diff
@@ -4,3 +4,7 @@
 - Example:
   - A
   - B
+- UxmPracticeTest:
```
sihugh, line 7: The thing I always need to check by seeing it running, is that this test name corresponds with the expected header in the app. I'm always 99% sure, but never quite 100%. 😰 
(It looks good to me FWIW, and checks out in the regenerated specs)

sihugh (commented):
This looks like it'll do the right stuff. There's just the question of whether you might want to split it into a couple of steps before going to production.

gclssvglx (approved):

PR #456, opened 2023-06-08
Add security.txt to the assets and bouncer services

sihugh:
This configuration redirects requests for the security.txt file to Cabinet Office's vulnerability disclosure program.

We want to add this feature to assets.publishing.service.gov.uk and all of the sites that bouncer caters for.

The configuration is identical to that for www.gov.uk, so I've extracted that as a partial and reused it across the services.

Bouncer had its stale cache grace time of 1 day specified in seconds, so I've tweaked that to `24h` so that it's the same as the other two.

I regenerated the specs with each commit to keep the config changes connected.

DGU's config is held in the govuk-aws repo, but that will be included in this work. 

https://trello.com/c/VLBjBTMe/313-put-securitytxt-on-bouncer-and-dgu

### DIFF
--- a/spec/test-outputs/assets-eks-integration.out.vcl
+++ b/spec/test-outputs/assets-eks-integration.out.vcl
@@ -81,6 +81,11 @@ sub vcl_recv {
      error 801 "Force SSL";
   }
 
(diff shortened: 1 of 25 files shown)


robinjam (approved):
LGTM 👍

PR #457, opened 2023-06-08
Make healthcheck probes more configurable.

sengi:
See alphagov/govuk-cdn-config-secrets#184 for the config change which goes alongside this.

While we're there, delete some disused config.

Easiest to review commit-by-commit.

https://trello.com/c/RkVL0rMF

Tested: already deployed (in integration and staging).


theseanything (approved):
Nice!

PR #459, opened 2023-06-23
Add bank hols ab test

chao-xian:
We think everything works on the new EKS based infrastructure but we want to double check everything all the way to production. This enables a very small AB Test on the Bank Hols page.

See https://github.com/alphagov/frontend/pull/3660

[Trello](https://github.com/alphagov/frontend/pull/3660)


BeckaL (approved):
I don't know loads about our A/B testing, but this roughly makes sense to me, so approving unless you'd like specific approval from someone who is more familiar

hannako (approved):

PR #463, opened 2023-07-03
ES 6.7 AA test config

Rosa-Fox:
[Trello](https://trello.com/c/AH3QERus/613-deploy-the-50-50-aa-test-for-the-search-team)

DO NOT MERGE! We need to wait for the [Corresponding rendering app PR](https://github.com/alphagov/finder-frontend/pull/3053) to be approved.

The purpose of the A/A test is to test the % of allocation of traffic/sample population based on the current version of Elastic Search (6.7) against itself. A (ES 6.7) vs A (ES6.7). The A/A is solely to identify how the approach allocates traffic. This will provide benchmark data to aid analysis in future AA tests.

[Corresponding rendering app PR](https://github.com/alphagov/finder-frontend/pull/3053)


leenagupte (approved):

configs/dictionaries/essixpointseven_percentages.yaml
```diff
@@ -0,0 +1,4 @@
+---
+A: 1
+B: 1
+Z: 98
```
sihugh, line 4: Should this be:

```
A: 50
B: 50
```
with no Z? Can chat off-thread

configs/dictionaries/ab_test_expiries.yaml
```diff
@@ -7,3 +7,4 @@
 ---
 Example: 86400
 BankHolidaysTest: 86400
+EsSixPointSeven: 604800
```
sihugh, line 10: Looks like a week 👍

configs/dictionaries/essixpointseven_percentages.yaml
Rosa-Fox: Thanks! Simon explained I need to remove the Z as it is only needed in the rendering app config, not the CDN config because A or B need to be assigned by the CDN.

PR #467, opened 2023-08-16
Reinstate ElasticSearch 6.7 AA test

Rosa-Fox:
[Trello](https://trello.com/c/VfqiGVjQ/318-reinstate-elasticsearch-67-aa-test-for-the-search-team)

[We recently ran an AA test for the Search team](https://github.com/alphagov/govuk-cdn-config/pull/463).

We are going to run it again, hence reinstating it in this PR!

The purpose of the A/A test is to test the % of allocation of traffic/sample population based on the current version of elastic (6.7) against itself. A (ES 6.7) vs A (ES6.7). The A/A is solely to identify how the approach allocates traffic. This will provide benchmark data to aid analysis in future AA tests.

This time around, the code will stay the same. We will use the same custom dimension and the variant split will still be 50/50. The only difference is that the custom dimension has been scoped as ‘user' as opposed to ‘session’ like it was last time. Scoping the custom dimension has already been done by a performance analyst.

We plan to deploy this on Thursday 17 August for 1 week.

[Corresponding rendering app PR](https://github.com/alphagov/finder-frontend/pull/3135)


gclssvglx (approved):
Looks great 👍

